PR #18, opened 2025-07-17
Update README.md

imcompiling:
docker was complaining about "enabled" not being able to be converted into a boolean. Correct readme to suggest correct value.


drallgood:
merged. thanks for catching that

Issue #24, opened 2025-08-03
Config.yaml Variables Ignored


zacaronii:
I am running into an issue where some of my config.yaml variables seem to be not registering. It's possible I'm missing something, although I've tried to check it over a few times with no luck finding what my error is, if so. 

I noticed that all books in my ABS library were being added to my want to read, even though I had the variable set to false. I then tried setting it to dry run mode, but it still did the same thing. I also noticed I don't see anything in the /app/data and /app/logs directories.

This is my most recent config file and docker setup in Unraid.

<img width="335" height="385" alt="Image" src="https://github.com/user-attachments/assets/ef724b06-1f3f-40ba-97b0-553e113bfb4b" />

```
logging:
  level: info        # debug, info, warn, error
  format: json       # json or text for improved readability during development
  pretty: false      # human-readable formatting for logs
  file: /app/logs/sync.log  # optional log file path

audiobookshelf:
  url: https://******
  token: *******

hardcover:
  token: ********
  rate_limit: 10     # requests per second
  rate_burst: 15     # burst capacity

sync:
  interval: 1h       # sync frequency (10m, 30m, 1h, etc)
  want_to_read: false # sync "Want to Read" status
  owned: false        # mark books as owned
  match_mode: continue # book matching behavior (continue, skip, fail)
  incremental: true # incremental sync strategy enabled (true, false)
  min_progress: 0.01   # minimum progress change to sync (0.01 = 1%)
  state_file: /app/data/sync_state.json # state storage location
  dry_run: true     # simulation mode without making changes
  force_full_sync: false # force full sync regardless of state
  libraries:         # library filtering (optional)
    include: ["Audiobooks"] # only sync these libraries (by name or ID)
```

drallgood:
Ah.. wait.. that syntax looks outdated.
I think you copied it from the README right? just saw it's wrong

drallgood:
@zacaronii for 2.1.0 use this file as template.

I will see that I clean this up for 3.0.0 (it's almost ready!)

zacaronii:
I used the updated config.yaml and reconfigured it. I found that the dry run mode was working as expected but even when I toggle off want to read it still appears to be adding everything in my ABS library to my want to read.

drallgood:
@zacaronii  thanks. will have a check in a bit... shouldn't be to hard to figure out what's wrong...

zacaronii:
[...]

PR #28, opened 2025-08-10
Fixing CONFIG_PATH var

therobbiedavis:
# Fixed:
- Fixed the example docker-compose. The example docker-compose uses CONFIG_FILE for the environment variable, however through my own trial and error I found this to be incorrect. The correct environment variable should be CONFIG_PATH.


therobbiedavis:
@drallgood From this [comment](https://github.com/drallgood/audiobookshelf-hardcover-sync/issues/24#issuecomment-3148537714), it looks like main is either being developed on directly or is not necessarily in line with the latest release/:latest docker image, which is what I tried. So feel free to dismiss the PR if that is the case.

drallgood (approved):

PR #57, opened 2025-10-31
Add JSON tags to BookNotFoundInfo fields

Clancey:
This fixes 2 issues. One is, if there is an ISBN number that fails to look up, the book will never sync since it doesn't fall back to ASIN or Tittle/Author. This fixes syncing for 60% of my audiobooks! The second issue is when an audiobook is not found, the web ui doesn't show the data: 
image


drallgood:
@Clancey  Thank you. I pulled it into the develop branch and will release it soon.

PR #152, opened 2026-07-10
fix: skip unread books before Hardcover lookup

matthewbrahms:
## Summary of Changes

Skips unstarted books before calling the Hardcover API instead of after to save throttling and API calls.

Previously, when `process_unread_books` was disabled, every library item was sync-checked as to whether or not the book had been started — making a lot of unneeded API calls and creating "book not matched" mismatch logging records for books that were about to be skipped. 

These changes build on the recent `enhanceBookProgressFromUserData` change on `develop:` since progress is already merged from `/api/me` early, the unread check can now run before any lookup.

Also: books skipped by the unread filter are no longer counted in the `BooksSynced` summary stat (they still count toward total books processed).

## Testing Instructions

- `go test ./...` — all tests pass, including three new/updated ones:
  - `TestProcessBookSkipsUnreadBeforeHardcoverLookup`: an unstarted book
    triggers no Hardcover search calls and doesn't count as synced.
  - `TestProcessBookMergesUserProgressBeforeUnreadSkip`: a book with no
    progress in the library payload but progress in `/me` is NOT skipped and
    reaches the Hardcover lookup (pins the merge-before-skip ordering).
  - `TestEnhanceBookProgressFromUserData`: media progress takes precedence
    over listening sessions; the session fallback preserves the library
    item's `StartedAt`.
- Manual check: set `process_unread_books: false`, enable debug logging, and
  run a sync against a library containing unstarted books. Unstarted books
  log "Skipping unstarted book before Hardcover lookup" with no preceding
  Hardcover search requests, and the summary no longer counts them as synced.

## Checklist
- [X] Tests pass locally
- [x] Tests pass on CI
- [X] Changelog is updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- **Unread Book Filtering Before Hardcover Lookup**: Skip unstarted books (when `process_unread_books` is disabled) before matching the book in Hardcover, instead of after. This prevents unnecessary Hardcover API requests and mismatch records for books that were going to be skipped anyway.
+- **Sync Summary Accuracy**: Books skipped by the unread filter are no longer counted in `BooksSynced` (they still count toward total books processed).
+
 ## [v3.3.1] - 2026-07-01
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - **Unread Book Filtering Before Hardcover Lookup**: Skip unstarted books (when `process_unread_books` is disabled) before matching the book in Hardcover, instead of after. This prevents unnecessary Hardcover API requests and mismatch records for books that were going to be skipped anyway.
 - **Sync Summary Accuracy**: Books skipped by the unread filter are no longer counted in `BooksSynced` (they still count toward total books processed).
+- Thanks to @matthewbrahms for this contribution.
 
 ## [v3.3.1] - 2026-07-01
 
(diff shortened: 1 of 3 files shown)


internal/sync/service.go
```diff
@@ -1064,6 +1064,19 @@ func (s *Service) processBook(ctx context.Context, book models.AudiobookshelfBoo
 		})
 	}
 
+	// Skip books that haven't been started unless ProcessUnreadBooks is true.
+	// This must happen before any Hardcover lookup so that a broad library pass
+	// doesn't spend API capacity (or create mismatch records) on books this
+	// profile has opted out of processing.
+	// Don't skip finished books even if CurrentTime is 0 (they have progress=1.0).
+	if book.Progress.CurrentTime <= 0 && !s.config.Sync.ProcessUnreadBooks && !book.Progress.IsFinished {
+		bookLog.Debug("Skipping unstarted book before Hardcover lookup", map[string]interface{}{
```
drallgood, line 1073: Nit (logging consistency): New log key is only "current_time". In the removed block, logging included book_id and title context. You may want to retain at least book_id for easier traceability in debug logs.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- **Unread Book Filtering Before Hardcover Lookup**: Skip unstarted books (when `process_unread_books` is disabled) before matching the book in Hardcover, instead of after. This prevents unnecessary Hardcover API requests and mismatch records for books that were going to be skipped anyway.
```
drallgood, line 11: Nit: Feel free to add your name as credit to the changelog

drallgood (dismissed):
Thanks for the contribution. Definitely a case that can be handled better! 

I have added 2 nits, just little things we could improved. They wouldn't block a merge though. Just let me know and I will merge as is.

internal/sync/service.go
matthewbrahms: Good call, thanks for catching that.  Definitely will include the `book_id`. 👍

CHANGELOG.md
matthewbrahms: Ok, will do.  Happy to contribute to the project.  ❤️ ABS and Hardcover and glad I found this project!

drallgood (approved):
Looks great! Thanks for the contribution!

Issue #156, opened 2026-07-18
Rereads not correctly detected


drallgood:
Currently rereads don't get detected because the logic relies on ABS's startedAt date which never changes

imcompiling:
My work around is to just remove previous progress from abs

drallgood:
That probably works too. I think I found a solution already (it's on develop) but am testing it still 
@imcompiling

PR #167, opened 2026-08-26
Fix infinite reread-split loop for in-progress books

### LINKED ISSUE #166: Infinite reread-split loop for any book left in progress across sync cycles
## Summary

`handleInProgressBook`'s stale-reread detection (added for #156) creates a new "reread" every single sync cycle for any book that stays in-progress across multiple runs — not just genuine rereads. Over ~2 weeks this generated 60-90+ duplicate `user_book_read` rows for two books in my library that I was simply still listening to.

## Root cause

`internal/sync/service.go`, two spots (~line 2846 and ~line 2974):

```go
if latestFinishedReadDate != "" && existingStartedAt <= latestFinishedReadDate {
    splitRereadFromStaleUnfinished = true
    // ...closes the current unfinished read with finished_at = today, creates a new one
}
```

`latestFinishedReadDate` is computed from all finished reads on the same edition. The first time this fires for a book, it closes the active read with `finished_at` = today's date and opens a fresh one starting today. On the *next* cycle, the new read's `started_at` is today, and the just-closed read's `finished_at` is also today — so `existingStartedAt <= latestFinishedReadDate` (today <= today) is true again. This repeats every sync interval for as long as the book stays in progress, since ABS naturally reports a slightly different `current_time` each cycle (satisfying the earlier progress-diff threshold that gates entry into this branch).

Observed in production logs: the same book generating a new `"Closed stale unfinished reread, creating a new active read"` entry every ~30 minutes for 13 days straight.

## Fix

Use strict `<` instead of `<=` at both sites. A read is only treated as a stale/stale-started reread when it genuinely started *before* the latest finished read on that edition — not when the dates are simply equal (which is exactly the state left behind by the previous cycle's own split). This preserves the original #156 fix for next-day-or-later rereads while breaking the same-day feedback loop.

PR: #(to be filled in) — branch `fix/reread-date-comparison-loop` on my fork, tested against a real deployment (built the patched binary, ran it against my actual Audiobookshelf + Hardcover accounts across several forced sync cycles) — confirmed the duplicate-read count stayed flat across multiple consecutive cycles where the unpatched version had been creating one duplicate per cycle.

Happy to open a PR if useful.

artic5693:
Fixes #166.

## Problem

The stale-reread detection added for #156 uses `existingStartedAt <= latestFinishedReadDate` to decide whether an in-progress read should be closed and split into a new "reread". Once it fires once, the newly-created read starts "today" and the read it just closed also finished "today" — so the same `<=` comparison matches again on the very next cycle. For any book that stays in progress across multiple sync runs (i.e. most books, most of the time), this creates a new duplicate `user_book_read` every single sync interval, forever.

I hit this in production: two books accumulated 60-90+ duplicate read entries over about two weeks of normal use, all stamped with `started_at`/`finished_at` clustered on just two calendar dates.

## Fix

Change both occurrences of the comparison from `<=` to `<`. A read is only spun off as a stale reread when it genuinely started strictly before the latest finished read on the same edition — same-day dates (which is what the previous cycle's own split leaves behind) no longer match, breaking the loop. The original #156 behavior (detecting a reread that starts on a later day) is unaffected.

## Testing

- `docker build` succeeds with the patch.
- Deployed the patched binary against my real Audiobookshelf + Hardcover accounts (replacing the `:latest` image in my own docker setup).
- Forced 4 consecutive sync cycles via `/api/sync` against the two previously-affected books. Read count stayed flat across all 4 cycles; the unpatched version had been producing one new duplicate read per cycle (every ~30 min) for the prior two weeks.


drallgood:
@artic5693 Could you please rebase the branch with develop? it has some changes already.

artic5693:
Rebased onto `develop` as requested — no conflicts (the two changes touch different parts of `handleInProgressBook`). Verified it still builds cleanly and the fix's comparison (`existingStartedAt < latestFinishedReadDate`) survived the rebase intact.

drallgood (approved):
A solid, minimal fix (two `<=` → `<` changes) that correctly breaks the infinite reread-split loop.

### Root Cause

The loop is real and the analysis is correct:

1. First split: old read closed with `finished_at = today`, new read created with `started_at = today`
2. Next cycle: `existingStartedAt (today) <= latestFinishedReadDate (today)` → matches → split again
3. Repeats forever for any book that stays in-progress

The existing guard at line 2834-2840 (skipping zero-progress reads where `StartedAt == FinishedAt`) doesn't catch this because the old split-closed read has `started_at = earlier date` and `finished_at = today` — they're not equal.

### Fix Analysis

Changing to `<` correctly distinguishes:
- A **genuine** stale reread: `started_at = Aug 25` < `latest_finished = Aug 26` → triggers
- A **self-inflicted** duplicate from a prior split: `started_at = Aug 26` < `latest_finished = Aug 26` → doesn't trigger

The edge case where someone genuinely starts a reread on the same calendar day as their last finished read is irrelevant — the read just keeps its existing `started_at`, which is correct.

PR #174, opened 2026-09-10
Limit raw API response logging to debug

Snuffy2:
## Summary

Routes the full Audiobookshelf library-items API response through debug logging instead of writing it directly to standard error. This keeps normal logs concise while preserving the raw response for troubleshooting when debug logging is enabled.

## What Changed

- Replaced the unconditional raw API response dump to standard error with a debug-level log entry.
- Preserved the complete response body in the debug log for diagnostics.

## Why

The direct standard-error write bypassed the configured log level and exposed a potentially large raw API payload during normal operation. Using the existing logger honors the application's logging configuration while retaining diagnostic visibility at debug level.


drallgood (dismissed):

drallgood:
Looks good. Could you please also update the changelog so you get credit? ;)

drallgood (approved):

PR #175, opened 2026-09-10
fix: prevent Hardcover mutations during dry runs

Snuffy2:
## Summary

Makes profile-level dry-run mode a hard safety boundary for Hardcover writes while preserving the read operations needed to preview a sync. It also exposes dry-run mode in Profile Options and clearly identifies active dry runs on the Sync Status page.

## What Changed

- Configure dry-run behavior directly on the Hardcover client and short-circuit every current mutation entry point, including the generic GraphQL mutation path.
- Continue performing Hardcover lookups and other read-only requests so dry-run output reflects the changes a real sync would attempt.
- Prevent dry runs from advancing or saving sync state, ensuring a later real sync still applies mutations that were only previewed.
- Add **Dry Run** to the create and edit Profile Options forms and persist the setting with each profile.
- Return the active profile's `dry_run` flag in sync status responses and display `Syncing (dry run)` without changing the underlying `syncing` state used for styling and cancellation.
- Add regression coverage for client mutation suppression, service configuration, and dry-run state handling.
- Document the fix in the changelog.

## Why

Existing dry-run checks covered only some sync paths, so finished-book processing, reading records, edition corrections, ownership updates, or future direct mutation calls could still change Hardcover. Guarding writes in the real Hardcover client establishes a single enforcement boundary while keeping lookup behavior representative.

Skipping state persistence is equally important: a successful no-op must not be recorded as applied, or the next real sync could incorrectly skip it.

## Validation

- `go test ./...`
- `go vet ./...`
- `go build ./...`
- `node --check web/static/app.js`
- Validated in a container run with `dry_run=true`: the status endpoint returned `status: "syncing"` and `dry_run: true`, the page served the updated label logic, mutation attempts were logged as skipped, and the container remained healthy.


drallgood:
The decorator is a legitimate pattern and it's hitting the right layer — blocking at the HardcoverClientInterface boundary is what actually guarantees every mutation gets caught, which is the core fix here. My hesitation isn't with the concept, it's with the failure mode: it embeds the whole interface, overrides a subset (8 of 24 methods), and silently falls through to the embedded client for anything else. That means a future mutation added to the interface would bypass dry-run without failing loudly — the exact class of bug this PR is fixing, just moved one hop downstream. The sentinel return values ("-1", 0, true, nil) also work today only because every caller happens to gate on the error and respect the -1 convention; there's no contract that keeps that true.

A simpler alternative worth considering: put the check inside the actual client as a DryRun field, with each mutation method doing if c.dryRun { log; return zero-value }. That gives us:

- No coverage problem — mutations are explicitly enumerated, reads untouched, no possible fall-through
- The existing findExistingUserBookForBook lookups at service.go:562 keep working without the unwrap dance
- Sentinel values live next to the real implementations, so they can't diverge
- Same one-line callers, same behavior
The cost is dry-run policy leaking into the transport client, but this client is app-specific and already carries endpoint/state logic, so it's mild.

I'd frame the decision as: if we want dry-run to extend to future mutations with zero discipline required, keep the decorator but make it fail-closed (log-and-skip anything it doesn't explicitly recognize, plus a test that enumerates all interface methods and asserts read/blocked classification). If the priority is least code and no parallel type to maintain, the in-client flag is the better fit.

I'm happy with either — I'd just rather we not ship the current fail-open version, since silent passthrough in dry-run is precisely the bug (#175) set out to kill.

Snuffy2:
Make sense. I'll refactor.

drallgood:
Thanks for the revision @Snuffy2 .
This addresses the structural concerns well. Moving the guard into the client and gating the generic GraphQLMutation entry point (client.go:467) closes the fail-open drift issue: any future mutation that routes through GraphQLMutation is blocked automatically, even if a per-method guard is forgotten. I confirmed the guarded set (the 8 per-method guards + the generic gate) covers every mutation on HardcoverClientInterface, and since the client is created per-profile (multiuser/service.go:341), SetDryRun can't affect other flows. The state-file guards in Sync() and the two processBook/HandleFinishedBook state writes remain correct, and the old decorator/interface-seam code is nicely gone.

A few minor nits, none blocking:

1. Sentinel return values (0, true, "-1") now live next to the real implementations, which is much safer — but a one-line comment tying CreateUserBook → "-1" to the existing convention in findOrCreateUserBookID (service.go:479) would prevent a future refactor from treating it as a real user-book ID.
2. Test intent: dry_run_test.go proves the client gate implicitly (a client built with no httpClient passing means no transport was attempted) — an explicit "no HTTP request" assertion would make that obvious to future readers.
3. Status badge: the dry-run label depends on DryRun surviving on the status record across mid-sync progress updates — worth a quick confirm that updates mutate the record in place rather than replacing it, otherwise the badge could drop mid-sync.

Approving from my side once those are settled.

Snuffy2:
1. Comment added to `findOrCreateUserBookID`
2. Added to dry_run_test.go
3. `StartSync` stores `DryRun` on the status record in [service.go line 264](https://github.com/drallgood/audiobookshelf-hardcover-sync/pull/175/files#diff-8b9c0166c233bcf631abfa739c09d88613ac0014f6800eaf80820dcc9a4bbc8dR264). `GetProfileStatus` retrieves that same record and updates its progress fields in place ([lines 165–233](https://github.com/Snuffy2/audiobookshelf-hardcover-sync/blob/fec0241b8cafeb9ff2a15f2b962b24db93a4baa3/internal/multiuser/service.go#L165-L233)), so `DryRun` is retained throughout the active sync. The record is replaced only when the sync leaves the `syncing` state, at which point the dry-run syncing label no longer applies.

drallgood:
Verified 097bcd3 and the -1 comment — both look good. The httptest-based assertion (no requests made for any of the 9 guarded paths) is a real improvement over the implicit version, and the findOrCreateUserBookID note removes the sentinel ambiguity. The GetProfileStatus in-place update explanation also confirms the badge stays accurate through the sync. 
Approving.

Thanks @Snuffy2 !!!

drallgood (approved):

PR #176, opened 2026-09-11
fix: correct rate limiter pacing and recovery

Snuffy2:
## Summary

Corrects Hardcover request pacing, concurrency, and recovery so syncs respond predictably to rate-limit guidance without appearing stalled or sending requests too quickly. The limiter now distinguishes normal pacing, temporary server-directed pauses, daily quota exhaustion, and fallback backoff.

## What Changed

- Pace sequential and concurrent requests consistently, including queued and canceled requests
- Hold concurrency permits for the full HTTP request and apply response guidance before admitting queued work
- Remove a duplicate rate-limit wait from publisher searches so each HTTP request is admitted and counted once
- Honor valid `Retry-After` values and quota reset windows as temporary pauses
- Wait until the reported reset when the daily quota is exhausted, with a separate 24-hour safety cap
- Apply bounded exponential backoff to bare, partial, or malformed 429 responses without shortening a longer existing pause
- Cap ordinary pauses and adaptive pacing at `DefaultMaxBackoff`
- Calculate sustainable pacing from IETF and legacy quota headers
- Restore the configured request rate when healthy quota data returns, while preserving elevated pacing when no quota data is available
- Count every 429 response once and clarify logs for throttling, adaptive pacing, recovery, and reset scheduling
- Remove the unused token-bucket implementation and obsolete burst setting from application configuration, examples, Helm values, and documentation
- Add migration guidance for removing `RATE_LIMIT_BURST` and `rate_limit.burst`
- Consolidate rate-limiter tests around observable pacing, concurrency, cancellation, backoff, reset, recovery, metrics, and logging behavior

## Why

The previous limiter could let concurrent requests accelerate the configured pace, release capacity before a request completed, and recover inconsistently after throttling. Some 429 responses could also bypass useful backoff or replace a longer server-directed pause with a shorter fallback delay.

These changes centralize request admission, keep all waits bounded, honor authoritative reset information, and recover only when the response provides evidence that normal pacing is safe.

## Validation

- `go test -race ./...`
- `go vet ./...`
- `go build ./...`
- `git diff --check`

## Checklist

- [x] Tests pass locally
- [x] Tests pass on CI
- [x] Changelog is updated

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -22,9 +22,6 @@ SHUTDOWN_TIMEOUT=10s
 # Minimum time between requests (30 requests per minute)
 RATE_LIMIT_RATE=2s
 
-# Maximum number of requests in a burst
-RATE_LIMIT_BURST=1
-
 # Maximum number of concurrent requests
 RATE_LIMIT_MAX_CONCURRENT=1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 - **Limit raw API response logging to debug**: Routes the full Audiobookshelf library-items API response through debug logging instead of writing it directly to standard error. (#174)
 - **Prevent Hardcover mutations during dry runs**: dry_run was not working correctly. Prevent dry-run syncs from calling Hardcover mutation APIs while preserving the read operations needed to report intended changes, and identify active dry runs on the Sync Status page. (#175)
-- **Correct rate limiter pacing and recovery**: Corrects Hardcover API pacing so rate-limit handling remains bounded, responds to authoritative server guidance, and recovers cleanly after throttling. This prevents concurrent waits from accelerating requests and prevents incomplete or unguided rate-limit responses from leaving syncs indefinitely stalled or incorrectly paced. (#176)
+- **Correct rate limiter pacing and recovery**: Corrects Hardcover API pacing so rate-limit handling remains bounded, responds to authoritative server guidance, and recovers cleanly after throttling. This prevents concurrent waits from accelerating requests and prevents incomplete or unguided rate-limit responses from leaving syncs indefinitely stalled or incorrectly paced. Removes the obsolete burst setting and unused token-bucket state; admission is governed by request pacing and concurrency. (#176)
 
 ## [v3.6.0] - 2026-09-01
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 - **Limit raw API response logging to debug**: Routes the full Audiobookshelf library-items API response through debug logging instead of writing it directly to standard error. (#174)
 - **Prevent Hardcover mutations during dry runs**: dry_run was not working correctly. Prevent dry-run syncs from calling Hardcover mutation APIs while preserving the read operations needed to report intended changes, and identify active dry runs on the Sync Status page. (#175)
+- **Correct rate limiter pacing and recovery**: Corrects Hardcover API pacing so rate-limit handling remains bounded, responds to authoritative server guidance, and recovers cleanly after throttling. This prevents concurrent waits from accelerating requests and prevents incomplete or unguided rate-limit responses from leaving syncs indefinitely stalled or incorrectly paced. (#176)
 
 ## [v3.6.0] - 2026-09-01
 
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -41,6 +41,8 @@ If you previously used environment variables for configuration, you will need to
 
 - **`HARDCOVER_SYNC_DELAY_MS`**: Replaced with request pacing and concurrency limiting
   - **Migration**: Use the rate limiting configuration in `config.yaml` instead
+- **`RATE_LIMIT_BURST` / `rate_limit.burst`**: This setting has been removed and no longer has any effect
+  - **Migration**: Delete it from your environment or YAML configuration. To adjust rate limiting, use `RATE_LIMIT_RATE` / `rate_limit.rate` to control how often requests are sent and `RATE_LIMIT_MAX_CONCURRENT` / `rate_limit.max_concurrent` to limit simultaneous requests.
 - **`AUDIOBOOK_MATCH_MODE`**: This legacy option has been removed
   - **Migration**: No action needed, improved matching is now the default
 
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -39,7 +39,7 @@ If you previously used environment variables for configuration, you will need to
 
 ### Removed Environment Variables
 
-- **`HARDCOVER_SYNC_DELAY_MS`**: Replaced with token bucket rate limiting
+- **`HARDCOVER_SYNC_DELAY_MS`**: Replaced with request pacing and concurrency limiting
   - **Migration**: Use the rate limiting configuration in `config.yaml` instead
 - **`AUDIOBOOK_MATCH_MODE`**: This legacy option has been removed
   - **Migration**: No action needed, improved matching is now the default
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -48,7 +48,6 @@ If you previously used environment variables for configuration, you will need to
 
 - **`HARDCOVER_BASE_URL`**: Override the Hardcover GraphQL API base URL (default: https://api.hardcover.app/v1/graphql)
 - **`RATE_LIMIT_RATE`**: Minimum time between Hardcover API requests (e.g., `2s` for 30 requests/minute)
-- **`RATE_LIMIT_BURST`**: Maximum burst size for requests (e.g., `2`)
 - **`RATE_LIMIT_MAX_CONCURRENT`**: Maximum number of concurrent requests (e.g., `3`)
 - **`CONFIG_FILE`**: Path to the configuration file (default: "./config.yaml")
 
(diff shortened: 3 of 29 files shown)


drallgood (changes requested):
Nice work on the pacing and wake-up logic — the admission-timestamp recheck loop, the scheduleChanged recycle pattern, and the authoritative-vs-fallback 429 classification are all sound and well-tested. Everything I checked statically holds up, including the cancellation race (no slot collapse) and the bounded backoff ceiling.

One design point I'd like to raise before merge:

Recovery resets the escalated pacing even when no quota data is present. Every 200 response that carries no IETF/legacy rate-limit headers flows through applyLegacyHeaders → setRate(r.minRate), so a single successful request wipes out all escalation from a previous bare 429. The summary says "restore the configured rate directly when healthy quota data returns," but the code restores it when there's no data. If the API intermittently returns headerless 429s, the client oscillates between the base rate and 8× base instead of converging to a sustainable pace, which defeats part of the backoff behavior this PR is fixing. TestRateLimiterFallsBackForBareTooManyRequests locks this in deliberately, but I'd suggest only restoring minRate when an actual authoritative/healthy header set is present, and leaving the escalated rate untouched when the response carries no rate-limit guidance.

A few smaller notes, all non-blocking:

1. OnRateLimit (rate_limiter.go:239) calls setBackoffUntil unconditionally and can shorten an existing authoritative deadline, unlike the guarded path in WithRateLimitHeaders. It's test-only in prod today, so not a live bug, but it contradicts the "preserve longer backoff" fix — same one-line guard would align it.
2. Daily-exhausted "pause" (rate_limiter.go:627-636) is capped at DefaultMaxBackoff, so a full-day exhaustion means one request every ~10 minutes, each instantly 429ing, until the daily reset. Bounded is clearly intentional, but the "Daily rate limit exhausted, pausing" message understates the repeated probing — worth acknowledging in the log.
3. Legacy exhausted log (rate_limiter.go:754-758) prints the uncapped reset duration while applyRetryAfter applies at most maxBackoff.
4. Nits: the expectPanic branch in TestNewRateLimiter is dead (always false); OnRateLimit's Retry-After branch skips drainBucket() unlike its sibling paths; and setRate/Wait log while holding r.mu, which is now hit more often with recovery-on-every-response — a slow logger would stall admission.

Snuffy2:
Sounds good, I'll let you know once it is ready for re-review

Snuffy2:
Thanks for the thorough review. The headerless success recovery issue was a good catch. I also found some additional dead code and refinements on further review; hence the expanded PR code.

## Addressed

- Successful responses without usable rate-limit headers now preserve the current escalated pacing. The configured rate is restored only when healthy quota data confirms that normal pacing is safe.

- Daily quota exhaustion now pauses until the reported reset instead of probing every ten minutes. Daily reset waits use a separate 24-hour safety cap, and reset values are bounded before duration conversion to prevent overflow.

- Legacy exhaustion logs now distinguish the reset reported by the server from the actual capped delay.

- Request admission is now centralized in the GraphQL execution path. Concurrency permits remain held for the full HTTP request, and response rate-limit guidance is applied before queued requests are admitted.

- Removed the redundant rate-limit admission in `SearchPublishers`, which previously added an extra pacing delay and counted a single request twice.

- Removed the unused token-bucket mechanism entirely, including `drainBucket`, token state, and the obsolete burst configuration. Migration guidance now tells users to remove `RATE_LIMIT_BURST` or `rate_limit.burst`.

- Removed the unused `expectPanic` test branch and consolidated the associated tests around observable behavior.

## Verified—no change needed

- `OnRateLimit` already uses the guarded backoff helper, so it cannot shorten an existing authoritative deadline.

## Deferred

- I left the non-blocking suggestion to move logging outside the rate-limiter mutex unchanged. Doing so safely would require restructuring the locked sections, and there is no evidence that logging currently delays request admission. I would prefer to address that separately if needed.

drallgood (changes requested):
Thanks for the thorough follow-up. I re-reviewed the updated head and verified each item:

**Confirmed addressed**
- **Headerless success recovery** — `applyLegacyHeaders` now early-returns when no `X-RateLimit-Remaining` is present, so a bare 429's escalated pacing survives a headerless 200; `TestRateLimiterFallsBackForBareTooManyRequests` now asserts exactly that (and that only healthy headers restore the base rate). Good.
- **Daily exhaustion** — `applyDailyResetWait` with the separate `DefaultMaxDailyResetWait` cap, bounded before Duration conversion, and the daily bucket taking precedence over `Retry-After` are all correct and well-tested at both unit and `httptest` level (`TestGraphQLQuery_DailyResetOverridesRetryAfter`).
- **Legacy logging** — the `reset_in` vs applied `backoff` split is in place.
- **Acquire/Release** — `TestGraphQLQuery_MaxConcurrentLimitsActiveRequests` proves the permit is now held for the full round-trip, and `SearchPublishers`' duplicate admission is gone (it now routes through `GraphQLQuery` only).
- **Token bucket / `expectPanic` / config removal** — all clean, including helm and migration docs; no stale references anywhere in code.

**One claim I have to push back on**

> OnRateLimit already uses the guarded backoff helper, so it cannot shorten an existing authoritative deadline.

It uses the *unguarded* variant. `OnRateLimit` calls `applyExponentialBackoff(false)` (rate_limiter.go:225), and the guard is disabled when `false`:

```v
if !preserveLongerPause || until.After(r.backoffUntil) {
    r.setBackoffUntil(until)
}
```

So `OnRateLimit(0)` while an authoritative deadline is set will unconditionally overwrite it — it can shorten the deadline, same as before. The `WithRateLimitHeaders` fallback path correctly passes `true` (rate_limiter.go:429). Since `OnRateLimit` is currently test-only this isn't a live bug, but the verification note is inaccurate; passing `true` is a one-word fix and would make the claim true.

**Minor, non-blocking**
- `enforceRateLimit` (and the admission-only `Wait` wrapper) now has no production callers — it survives only for `TestClient_enforceRateLimit`. Given this PR's explicit dead-code cleanup (token bucket, `expectPanic`), it seems like it should go too.
- Holding a permit for the full request duration is the right call, but `executeGraphQLOperation` builds a fresh `http.Client{}` with no `Timeout` (client.go:483). With the default `max_concurrent: 1`, one hung connection now holds the only permit indefinitely and stalls the whole client — previously a hang only froze its own goroutine. A transport `Timeout` (or depending on the caller's ctx deadline) is worth adding.
- The "reset values are bounded before duration conversion" guarantee holds only in `applyDailyResetWait`; the near-exhausted daily branch still computes `time.Duration(reset)*time.Second` (rate_limiter.go:614) and the per-window pause (rate_limiter.go:656) without a cap. Absurd values degrade to the fallback branch rather than pausing rather than crash, and realistic resets (≤86400s) are safe, so this is cosmetic — but it does undermine the stated overflow rationale if you want to be strict.

All substantive behavioral fixes check out. Happy to approve once the `OnRateLimit` guard is flipped to `true` (or the note corrected) and `enforceRateLimit` is either removed or `//nolint`-justified.

Snuffy2:
Thanks for the careful re-review. I’ve addressed the remaining points:

- Updated `OnRateLimit(0)` so fallback backoff cannot shorten an existing authoritative pause.
- Removed the unused `enforceRateLimit` method and admission-only `RateLimiter.Wait` API, including obsolete test coverage.
- Updated GraphQL execution to preserve the configured HTTP client timeout and custom transport while ensuring permits are released after timeout or request errors.
- Bounded the remaining header-derived reset and window calculations before duration conversion to prevent overflow.
- Added regression coverage for authoritative-pause preservation, HTTP timeout and permit reuse, and extreme header values.
- Simplified the related fallback calculations by removing a redundant flag, parsing legacy reset timestamps once, and reusing the bounded-duration helper.

drallgood (approved):
Verified everything on the final head, and this time I also built and ran the suite (`go build ./...`, `go vet ./...`, `go test -race ./...`, and 10× race-enabled runs of the timing-sensitive tests — all green). Both rounds' items check out:

**Confirmed**
- **`OnRateLimit` guard** — `applyExponentialBackoff()` now drops the flag and always uses `if until.After(r.backoffUntil)` (rate_limiter.go:790-797), so the fallback cannot shorten an authoritative pause. `TestRateLimiter_OnRateLimitFallbackPreservesLongerAuthoritativePause` covers exactly the failure mode I flagged. Solid.
- **Dead code removal** — `enforceRateLimit` and `Wait` are gone, including the `TestClient_enforceRateLimit` test; remaining test call sites migrated to an `acquireAndRelease` helper. No leftover references anywhere.
- **HTTP timeout preservation** — actually a bigger win than stated: `c.httpClient` was previously *unused* for requests, and `executeGraphQLOperation` built a fresh `http.Client` with no `Timeout`, so the configured `DefaultTimeout` (30s) never applied to the GraphQL path. Copying `c.httpClient` and wrapping only the transport now gives the request path its configured timeout, redirects, and transport. `TestGraphQLQuery_UsesConfiguredTimeoutAndReleasesPermit` proves both the timeout aborts a stalled request and the permit is released for reuse afterward.
- **Bounded header math** — `boundedSecondsDuration`/`boundedSecondsRate` order the overflow checks before duration conversion and are applied to the daily recovery window, per-window pause, sustainable rate, and `applyDailyResetWait`. `TestRateLimiterBoundsExtremeNonExhaustedIETFValues` exercises max-int and malformed values and confirms clamping to `DefaultMaxBackoff`. The legacy reset timestamp is parsed once and reused. Clean.
- **Permit lifecycle** in `executeGraphQLOperation` — every path after `Acquire` releases exactly once; confirmed no leaks or double-releases.

Nothing left blocking from me. I'll approve this now. The mutex-hold logging remains the one thing I'd eventually like to see revisited, but it's a separate cleanup and not a regression — agreed that it doesn't belong in this PR.

PR #178, opened 2026-09-12
fix(web): prevent status loader error from stray Hardcover block

Snuffy2:
## Summary of Changes

Remove a misplaced Hardcover mismatch-display block from the Sync Status loader. It referenced variables that do not exist in that method, causing successful profile status requests to log a JavaScript error. Add a changelog entry for the fix.

## Testing Instructions

- `node --check web/static/app.js`, `make test`, `make build`, and `git diff --check upstream/develop...HEAD` passed locally.
- `make lint` could not complete locally: installed golangci-lint v1.64.8 cannot decode export data produced by the current Go toolchain. CI quality checks remain pending.
- To verify in the UI, open Sync Status with a profile whose status endpoint succeeds and confirm it loads without an undefined-variable error in the browser console.
- No README change is needed for this error fix; the changelog records it.

## Checklist

- [x] Tests pass locally
- [x] Tests pass on CI
- [x] Changelog is updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **Sync Status fetch errors**: Remove misplaced Hardcover mismatch-display code from status loading so successful profile status requests no longer log an undefined-variable error.
 - **Limit raw API response logging to debug**: Routes the full Audiobookshelf library-items API response through debug logging instead of writing it directly to standard error. (#174)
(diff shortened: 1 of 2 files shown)


drallgood (approved):
**Review**: The fix is correct. Verified that `loadStatuses` never has `source`, `cleanData`, `details`, or `metadata` in scope (they only exist inside `extractBookData` and `renderBookDetails`), so the stray block threw a `ReferenceError` on every successful status fetch, caught by the per-profile `catch` and logged as `Error fetching status for profile ...`.

The removed code is also dead/stale rather than functional: it was an older variant of the HC block that already lives, correctly scoped and more complete (includes `data.id` in the match check, current edition-warning UI, and HC rating/slug/series/genres rows), in `renderBookDetails`. No coverage is lost.

PR #179, opened 2026-09-13
feat: checkpoint incremental state after each book

Snuffy2:
## Summary of Changes

Checkpoint changed incremental sync state after each processed book so completed work survives an interrupted library pass. Each run still scans from the first book and uses saved state to skip unchanged entries; that unchanged-book skip already existed after completed runs.

- Persist changed state through a temporary-file write, file sync, atomic rename, and directory sync where supported, while following configured state symlinks.
- Keep dry runs non-persistent, retain the final end-of-run save, and fail the sync when a checkpoint or final save fails.
- Intentionally treat a failed required Hardcover status update as a per-book failure: log it, leave that book uncheckpointed, and continue processing other books. Checkpointing after only the read update would falsely mark the requested status transition complete. If the status API keeps failing, the same book will fail again on later runs even when its read progress was already saved. Retry tests cover recovery without duplicating an already-applied read mutation.
- Keep blank-read cleanup best-effort: cleanup failures are logged but do not block a checkpoint after required read and status work succeeds.
- Keep an existing read with unavailable Hardcover status uncheckpointed and refetch the status on the next attempt. For a newly inserted read whose status was omitted, set the status appropriate to its progress. Progress skipped because its difference is below the update threshold remains eligible for a later retry, even if status reconciliation succeeds. Regression tests cover these paths, checkpointing, cancellation, durability, and retries. Update the README and changelog.

## Testing Instructions

Validated locally with Go 1.26.7: uncached `go test -race ./... -count=1`, `make lint`, `go vet ./...`, `go build ./...`, and `git diff --check upstream/develop...HEAD`. The upstream and fork Go 1.26 test/lint jobs, other test jobs, and security scans passed at this head.

## Checklist

- [x] Tests pass locally
- [x] Tests pass on CI
- [x] Changelog is updated


drallgood (changes requested):
**Re: the deleted status-mutation block after the read update (~old lines 2907ff, now `reconcileBookStatus()`)** — I think I've traced all three behavioral deltas and want to confirm each is intended:

1. **Failed status updates are now fatal for the book.** Previously `UpdateUserBookStatus(3)` / `(2)` failures were logged (`Error`/`Warn`) and the book still counted as processed with state advanced. Now `reconcileBookStatus()` returns the error, so the book fails, is not counted, and is retried on the next run. That's the right semantics given per-book checkpointing, and it's safe because the read progress update is idempotent. But the flip side is: a flaky status API now makes the *same* book fail on *every* run (logged as "Failed to process item") even though its progress is being synced fine — where old code would sync progress and just lag the status. Worth stating the intent explicitly in the PR description/CHANGELOG.

2. **The `BookStatusID > 0` guard drops the old best-effort transition for unknown status.** For responses that omit `bookStatusId` (0), the old code still issued IN_PROGRESS (unfinished) or COMPLETED (finished ABS); the new `statusNeedsReconcile()` guard treats 0 as "no evidence, take no action." I agree with the reasoning (never flip a status based on a missing field), and `CreateUserBook` sets a concrete status so fresh user books aren't affected — but this is a genuine behavior change: a user book whose *stored* status went stale won't be repaired by a progress update anymore in that (rare) 0-status response case. Confirm that's intentional.

3. **No-op books now persist state, so the status-repair no longer "eventually runs".** Combined with the verified-no-op advances: a no-op book (progress unchanged, status matching) is now written to state and skipped by `NeedsSync` on later runs. Previously it was re-verified every run, so a manual HC-side change to its reads/status would be reconciled on the next run automatically. Now it's only reconciled after ABS progress changes. If "HC is source of truth for post-sync manual edits" is the desired contract, the per-book `NeedsSync` skip is fine — just want to make sure the duplicate/manual-read cleanup (old `else if !isFinishedInHC` path, `deleteBlankReads`) lapsing for these books is accepted.

The rest of the refactor is solid: state written only after every mutation succeeds, `lastProgressUpdates` cache ordered after reconcile so a failed transition can't suppress the retry, and the `isFinishedInHC` `""` fix. Non-blocking; approving pending your answer on the three points above.

Snuffy2:
Thanks for the careful review. I’ve addressed your points in order:

1. I updated the changelog and PR description to clarify that a failed Hardcover status update leaves the book uncheckpointed for retry, even if its read progress was saved. A persistent failure may recur, while other books continue processing.

2. I fixed the existing-read case where Hardcover returns an unavailable status (`0`): after a successful read update, sync state remains unadvanced and the cached unknown status is invalidated so the next attempt fetches it again. I added a regression test. When sync creates a new read with an unavailable status, it now sets the status appropriate to the book’s progress rather than always setting `IN_PROGRESS`.

3. I need to correct my earlier characterization: before this PR, Incremental Sync already skipped unchanged books before contacting Hardcover when it had state saved from a completed run. Per-book checkpoints extend that behavior to books completed before an interrupted run. I also fixed the below-threshold case so progress that was *not* sent to Hardcover is not checkpointed and remains retryable. Hardcover-only manual changes still won’t trigger periodic reconciliation; I accept that existing limitation for this PR and would treat reconciliation as a separate feature.

drallgood (approved):
Thanks — I re-verified all three points against the updated head.

1. Retryable-status doc — the changelog entry (moved into Changed) is accurate and precise, including "an unavailable status on an existing read is not blindly changed or checkpointed after a read update."

2. Status-0 handling — confirmed on the code: the top-of-function if hcBook != nil && hcBook.BookStatusID == 0 { s.userBookCache.InvalidateByUserBook(...) } guarantees a fresh fetch on the next attempt (including across runs, since the user-book cache is disk-persisted with a 6h TTL — invalidating matters there). The read-update path correctly leaves state unadvanced, clears the progress guard, and returns success; the create path now uses desiredStatusID instead of hardcoding IN_PROGRESS. The two regression tests (UnavailableStatusAfterReadUpdateRemainsRetryable, FinishedNewReadWithUnavailableStatusUsesFinishedStatus) plus the table-driven MatchingReadNoOpStateHandling / MatchingFinishedReadNoOpAdvancesState cases pin exactly the contracts I was worried about.

3. Below-threshold retryability — handleSkippedProgressUpdate(true/false) splitting identical-progress (verified no-op, checkpoint) from below-threshold (not sent, stays retryable) is correct, and I confirmed the "incremental already skipped unchanged books pre-PR" characterization is right — per-book checkpoints really do just extend that to books completed before interruption.

Approve — the three concerns are resolved and well-tested.

PR #180, opened 2026-09-13
fix(web): make Sync Status refresh unobtrusive

Snuffy2:
#### Step 1 of 4 to Improve Sync Status UI

Make the five-second Sync Status refresh quiet: background polls leave the loading overlay hidden and update only cards whose status changed. Failed initial profile loads recover automatically with capped backoff, while Start and Cancel request failures remain visible on the affected card without changing the server-reported sync state.

- Prevent overlapping polls and stale responses from replacing a newer status snapshot; bound profile, status, and summary requests so a stalled request does not block later refreshes.
- Keep the last known profile status when a status request fails, label cached cards as last known data until recovery, and redirect to login when a current status request returns 401. Retry an initial profile-load failure from 5 seconds up to a 60-second cap; a valid empty list or authentication redirect ends that retry path.
- Reload authoritative status after Start or Cancel; show request failures as dismissible per-card action errors rather than fabricated sync errors.
- Preserve unchanged status cards and update relative timestamps during polling. Use delegated profile actions and escape profile identifiers in HTML attributes.
- Document the UI's browser minimums in the README and guard status loading when `AbortController` is unavailable.

## Testing Instructions

- `node --check web/static/app.js` and diff checks passed.
- Temporary, uncommitted JavaScript checks covered stalled and failed requests, silent recovery, empty profiles, authentication redirects, non-overlapping polls, the missing-`AbortController` path, per-card action errors, retry backoff/reset, cached-card unavailable/recovery, and relative-time rendering.
- `make test` and `make build` passed during the original local validation. Later changes are confined to the web UI and documentation; the current PR head's Test, Go 1.26, and scan checks passed on CI.

## Multi-Step Project

1. **Quiet background refresh (this PR):** Keep timer-driven polls silent and bounded, reject stale results, retain unchanged cards, recover from temporary profile-load failures, and show action failures on the relevant card.
2. **Per-run outcomes and live status API:** Give every attempted book one outcome, reconcile category counts to processed, classify technical failures separately, and publish immediate per-profile mismatch and missing-book results. Keep legacy fields usable while the UI moves to the new contract.
3. **Sync Status and View Details redesign:** Consume the outcome contract, remove the duplicate `Books Synced` and misleading all-found text, show the full category breakdown, and refresh open details smoothly. Preserve scroll and focus; when a card is replaced, move focus to Start/Cancel only if that was the previously focused action.
4. **Run lifecycle and report history:** Add phase/activity state, truthful canceled/failed partial reports, distinct last-attempted and last-successful timestamps, and a bounded persisted final report.

## Checklist

- [x] Tests pass locally
- [x] Tests pass on CI
- [x] Changelog is updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **Make sync status refresh unobtrusive**: Make Sync Status polling quiet and race safe so background refreshes do not flash the page or disturb the user's current view. By @Snuffy2.
 - **Sync Status fetch errors**: Remove misplaced Hardcover mismatch-display code from status loading so successful profile status requests no longer log an undefined-variable error by @Snuffy2. (#178)
 - **Limit raw API response logging to debug**: Routes the full Audiobookshelf library-items API response through debug logging instead of writing it directly to standard error. (#174)
 - **Prevent Hardcover mutations during dry runs**: dry_run was not working correctly. Prevent dry-run syncs from calling Hardcover mutation APIs while preserving the read operations needed to report intended changes, and identify active dry runs on the Sync Status page. (#175)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- **Make sync status refresh unobtrusive**: Make Sync Status polling quiet and race safe so background refreshes do not flash the page or disturb the user's current view. Automatically retry a failed initial profile load without requiring a manual refresh. By @Snuffy2.
+- **Make sync status refresh unobtrusive**: Make Sync Status polling quiet and race safe so background refreshes do not flash the page or disturb the user's current view. Automatically retry a failed initial profile load without requiring a manual refresh. By @Snuffy2. (#180)
 - **Sync Status fetch errors**: Remove misplaced Hardcover mismatch-display code from status loading so successful profile status requests no longer log an undefined-variable error by @Snuffy2. (#178)
 - **Limit raw API response logging to debug**: Routes the full Audiobookshelf library-items API response through debug logging instead of writing it directly to standard error. (#174)
 - **Prevent Hardcover mutations during dry runs**: dry_run was not working correctly. Prevent dry-run syncs from calling Hardcover mutation APIs while preserving the read operations needed to report intended changes, and identify active dry runs on the Sync Status page. (#175)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- **Make sync status refresh unobtrusive**: Make Sync Status polling quiet and race safe so background refreshes do not flash the page or disturb the user's current view. By @Snuffy2.
+- **Make sync status refresh unobtrusive**: Make Sync Status polling quiet and race safe so background refreshes do not flash the page or disturb the user's current view. Automatically retry a failed initial profile load without requiring a manual refresh. By @Snuffy2.
 - **Sync Status fetch errors**: Remove misplaced Hardcover mismatch-display code from status loading so successful profile status requests no longer log an undefined-variable error by @Snuffy2. (#178)
 - **Limit raw API response logging to debug**: Routes the full Audiobookshelf library-items API response through debug logging instead of writing it directly to standard error. (#174)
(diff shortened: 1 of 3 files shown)


drallgood (changes requested):
**Reviewed the change — this is a solid step 1.** The stalled-request recovery, sequence-guarded publishing, and the migration away from inline `onclick` handlers are all well done, and the "reload authoritative status after actions" approach is the right call. A few things worth addressing:

1. **`AbortController` needs a fallback or a documented minimum browser.** Safari < 17.2, Chrome < 121, and Firefox < 125 don't have it, and here a missing `AbortController` throws on the *first* profile/status request, which breaks status loading entirely rather than degrading gracefully. A small guard (e.g., skip the cancel/timeout wiring when `typeof AbortController === 'undefined'`) or a documented browser floor would cover this. This is the only item I'd consider blocking.

2. **Action failures are now toast-only.** Dropping the fabricated `status: 'error'` card state is defensible, but after a failed start/cancel the card shows nothing new until the next poll — and auto-refresh only runs while the Sync tab is active. If the user navigates away, there's no card-level feedback about the failure. A targeted status refresh (or an explicit per-card failure state) on the action error path would close that gap.

3. **Minor: the retry has no backoff.** A persistently failing profile endpoint gets polled every ~5–15s indefinitely. A capped exponential backoff for the `profileLoadFailed` retry would avoid hammering a down endpoint.

4. **Nit:** `data-sync-timestamp="${this.escapeHtml(lastSync)}"` is in a quoted attribute but uses `escapeHtml` (which doesn't escape quotes); for consistency with the new `escapeHtmlAttribute` hardening, use that here too. Low risk since it's server-generated, but it undercuts the fix.

Also noting: `renderStatuses` focuses the first start/cancel button when the previously focused element isn't an action button — only refocusing for actual action buttons would be less surprising. Not blocking.

Nothing here requires a redesign — just the `AbortController` decision and the action-failure feedback are the ones I'd want resolved (one before merge, one can be a follow-up).

Snuffy2:
Thanks for the detailed review. I’ve addressed the four numbered points:

1. Status loading now works when `AbortController` is unavailable; cancellation and timeout wiring are skipped in that case. I documented the UI’s browser minimums in the README. One correction to the version claim: [`AbortController` was supported earlier](https://github.com/mdn/browser-compat-data/blob/main/api/AbortController.json): Chrome 66, Firefox 57, Safari 12.1, and Edge 16. Other pre-existing UI features require newer browser versions.

2. Failed Start and Cancel requests now show a dismissible error on the affected card as well as a toast. The card retains the server-reported sync state, and a successful action clears the error.

3. Failed initial profile loads retry with capped exponential backoff, from 5 to 60 seconds. A successful load resets the backoff.

4. I removed the unused `data-sync-timestamp` attribute entirely; relative-time text and its tooltip still update.

A subsequent review also found that a failed status refresh could leave a cached card looking current. It now identifies the status as last known data until a successful refresh, and a current 401 response redirects to login.

I’ve left the additional focus behavior for Step 3, as planned. JavaScript syntax and temporary behavior checks passed, and the current PR head’s test and scan checks passed on CI.

drallgood (approved):
**Thanks — I re-reviewed the amended head and can confirm all of it lands:**

1. **AbortController fallback confirmed.** `fetchJsonWithTimeout` bypasses the timeout/cancel wiring when unavailable, `loadStatuses` falls back to sequence-only guarding, and the README documents browser floors (which are driven by pre-existing features like optional chaining and flexbox `gap`, not `AbortController`). Thanks for the compat correction. The tradeoff to keep in mind: on legacy browsers the fallback loses the per-request timeout, so a hung endpoint can delay refreshes there indefinitely — acceptable given it's documented and those browsers are already below the stated floor.

2. **Per-card action errors confirmed.** `actionErrors` (with dismiss), the card retaining the server-reported status, and clearing on a successful action all check out, and the `role="alert"`/`role="status"` usage is a nice touch.

3. **Backoff confirmed.** 5 → 10 → 20 → 40 → 60s (capped), gated by `nextProfileRetryAt`, reset on success/auth. Matches the claim exactly.

4. **`data-sync-timestamp` removal confirmed**, and the last-known-data banner + 401→login redirect on a status refresh are both present and correctly guarded by the request sequence.

Remaining nits — none of these block merge:

- **Backoff only covers initial profile-load failures**, since it's keyed off `users.length === 0`. If status/summary endpoints fail persistently (or profiles fail *after* a first successful load), polling stays at a fixed 5s with no backoff. The 15s request timeout bounds each attempt, so this is a nicety, not a bug.
- **Stale `actionErrors` entries** survive profile deletion (keyed by profile ID, never cleaned up). Invisible in the UI since the card no longer exists, but easy to sweep in the delete handler.
- The two **`console.log` debug lines** in `renderStatuses`/`fetchSyncSummary` are still there. Consider gating behind a debug flag when you touch this again.
- **Focus-restore behavior** deferred to Step 3 is fine by me — just keep it tracked.

Good to go from my side. The last-known-data handling in particular closes the gap the subsequent review found.

PR #183, opened 2026-09-14
feat(sync): account for per-run book outcomes

Snuffy2:
#### Step 2a of 4 to Improve Sync Status UI

Account for each attempted Audiobookshelf book with one per-run outcome so processed totals reconcile with the outcome categories. This prepares the backend for a subsequent live-status change while preserving the existing status API.

Technical lookup failures remain failures even when a title search finds a candidate; dry runs report planned work without persisting sync progress or mutating Hardcover.

## Testing Instructions

- `GOTOOLCHAIN=go1.26.7 make all` passed locally at `b5ed8c3` before the branch rewrite (race-enabled core tests, lint, and build).
- At the current PR head `3ed3469`, the Go 1.26 tests with race/coverage and lint, Trivy, container scan, and filesystem scan passed.
- Focused service tests cover reconciled outcomes, dry-run safety, failed lookups, candidate counting, and stale reread handling.
- No README change is needed because step 2a adds internal accounting without new API fields or configuration. The changelog records this step; step 2b will document the published live-status contract.

## Multi-Step Project

2a. Per-run outcome accounting (this PR): Give every attempted book one outcome, reconcile category counts to processed, and distinguish technical failures from conclusive no-results. Preserve dry-run and sync-state safety, and keep the existing UI and API usable without step 2b.

2b. Live per-profile status API: Publish mismatch and missing-book results as attempts finish. Expose a consistent current-run snapshot through status and summary responses, isolate profiles, and retain legacy response fields while the UI transitions.

3. Sync Status and View Details redesign: Consume the outcome snapshot, remove duplicate Books Synced and misleading all-found text, and show the full category breakdown. Fetch large detail lists on a run-scoped, on-demand path and refresh open details without losing scroll, selection, or focus. Restore focus to Start/Cancel only when that action previously held focus. Secure profile IDs across UI URLs and backend routes, clear deleted profiles’ action errors, and remove obsolete UI fallbacks and routine debug output while preserving public API compatibility.

4. Run lifecycle and report history: Show run phase and activity, retain truthful partial reports for canceled and failed runs, separate last-attempted from last-successful timestamps, and persist a bounded final report across restarts. Acknowledge Start Sync only after a run is accepted and queued; keep dry runs and legacy terminal status consistent with the actual result. Add paged or incremental details only if measured list size or polling cost warrants it.

## Checklist

- [x] Tests pass locally
- [x] Tests pass on CI
- [x] Changelog is updated


drallgood (changes requested):
### Requested changes before merge

Great work on the outcome accounting — I validated locally (`go build`, `go vet`, `go test -race ./internal/...` all green) and the reconciliation approach is solid. I have a few items before this merges.

**1. Don't drop the summary when a run ends with errors (required).**
Right now `Sync` returns an error when `GetUserProgress` fails or any library pre-count/fetch fails, and the multiuser layer responds by marking the profile `"error"` and discarding `BooksTotal`/`BooksSynced`/`BooksNotFound`/`Mismatches` entirely (`internal/multiuser/service.go`). A transient ABS hiccup now wipes out all visible counts for the profile. I want the partial summary still published: keep the run-level error handling, but always populate the summary on the profile status so users see partial progress.

**2. Keep identifier-lookup failures with title candidates visible (required).**
Committing `059c151` means books whose ASIN/ISBN lookup fails but where the title search finds a candidate are no longer recorded in `Mismatches` or `BooksNotFound`, and the outcome store isn't published until step 2b — so between now and 2b these books disappear from every user-visible surface. That's not acceptable. Please keep these books in the mismatch report (with the candidate's Hardcover details) so nothing regresses, and we can sort out the exact outcome semantics when 2b ships.

**3. Don't abort the entire run on an empty library item ID (required).**
`processLibrary` returning `errInvalidLibraryItemID` and `Sync` propagating it to `return err` is too aggressive for a single malformed item — it skips the mismatch save, state save, and cache saves, and flips the whole profile to error. Treat it like the other recoverable per-item cases: log and skip/record it, then continue the run.

**4. Fine as-is.**
Hard-failing the book when `GetUserBookReads`/`GetUserBook` errors instead of creating synthetic reads is the right call — keep it.

**5. Keep, but add a test and document it.**
`handleInProgressBook` now calls `reconcileBookStatus()` in the "read already finished with the same date" branch (previously a pure no-op). I'm fine with the mutation itself — it's self-limiting (only fires while `BookStatusID` is a positive, non-`FINISHED` value) and consistent with fixing stale statuses — but it's an unannounced behavior change. Please add a test covering the status-reconcile trigger in that branch and mention the behavior in the PR description.

**Nits (nice to have):**
- The `cachedBook == nil` "negative cache" branch in `findBookInHardcover` is dead code now that `getASINFromCache` never reports a nil hit — remove the branch and comment. (The behavior itself — treating stale nil entries as misses — is correct.)
- The raw `fmt.Errorf("found by title/author only")` in one branch of `findBookInHardcover` should use the typed `errHardcoverTitleOnly` sentinel instead of relying on `strings.Contains` in `processBook`.
- A test asserting `outcomeCounts.Total() == TotalBooksProcessed` for a full library run would lock in the headline reconciliation invariant.
- The PR description says "documentation and changelog are unchanged," but the head commit adds a CHANGELOG entry — please refresh the description.

Snuffy2:
Thanks for the detailed review. I’ve pushed the step 2a updates and addressed each point:

1. **Partial summaries:** A completed run with a run-level error still marks the profile as `error` while retaining its available counts, missing books, and mismatches. During the current service lifetime, the post-run summary reports the number actually processed.
2. **Title candidates after identifier failures:** These remain technical failures and appear in the legacy mismatch report with the returned candidate’s available Hardcover details.
3. **Empty item IDs:** A malformed item is logged and skipped, allowing later items and normal run finalization to proceed.
4. **Hardcover read errors:** I kept the hard-failure behavior. An incomplete remote read snapshot does not trigger synthetic read creation.
5. **Same-date finished reads:** This case takes the earlier finished-read branch, which already reconciles stale book status. I removed the redundant call from the unreachable later branch. The existing status-reconciliation retry test covers the earlier path, and I clarified it in the PR description.
6. **Negative-cache nit:** Removed the dead nil-cache branch; legacy nil entries are treated as misses so the lookup can retry.
7. **Title-only error nit:** Production code now uses the `errHardcoverTitleOnly` sentinel and `errors.Is` instead of the raw error and string matching.
8. **Count-invariant nit:** Added a full-library test containing a malformed item. It verifies that the malformed item is skipped and that the two attempted items produce matching processed and outcome totals.
9. **PR-description nit:** Updated it to acknowledge the changelog entry.

drallgood (approved):
Thanks — I went through the updated diff and each point checks out:

1. Partial summaries now survive run-level errors (summary populates regardless of `Status`).
2. Technical failures keep their `failed` outcome while the title candidate reappears in the mismatch report — confirmed in the test.
3. Missing item IDs warn-and-skip without aborting the run.
4. Read-fetch failures stay hard failures.
5. Removing the redundant reconcile call from the same-date branch is correct — that path is unreachable and the earlier finished-read branch already handles reconciliation.
6–9. Sentinels, dead-code cleanup, the malformed-item reconcile test, and the description all look good.

Only minor note for a follow-up: the failure-path mismatch now relies on `SearchBooks` returning the slug directly (the `GetBookByID` hydration was dropped), so verify the live API provides it — otherwise `hardcover_slug` will be blank on those rows. No blocker. Approving.

PR #184, opened 2026-09-15
Expose coherent per-profile sync status snapshots

Snuffy2:
#### Step 2b of 4 to Improve Sync Status UI

Expose coherent, per-profile current-run sync snapshots through the authenticated status and summary APIs. The snapshots track seven mutually exclusive book outcomes in real time while preserving legacy response fields and keeping the public aggregate status endpoint scalar-only.

## Testing Instructions

- `make test` — passed with the race detector and atomic coverage.
- `make build` — passed for the main binary and supporting commands.
- `make lint` — could not complete locally because the cached golangci-lint analyzer cannot decode the Go 1.27.1 export-data version; CI should run the workflow-installed current linter with the repository's configured Go toolchain.

## Multi-Step Project

2b. Live per-profile status API (this PR): Publish needs-review, not-found, and failed attention records as attempts finish. Expose a coherent run-scoped snapshot through authenticated per-profile status and summary responses, isolate profiles, and retain legacy response fields so the existing UI remains usable until step 3.

3. Sync Status and View Details redesign: Consume the outcome snapshot, remove duplicate Books Synced and misleading all-found text, and show the full category breakdown. Fetch large detail lists through a run-scoped, on-demand path and refresh open details without losing scroll, selection, or focus. Restore focus to Start/Cancel only when that action previously held focus. Secure profile IDs across UI URLs and backend routes, clear deleted profiles’ action errors, and remove obsolete UI fallbacks and routine debug output while preserving public API compatibility.

4. Run lifecycle and report history: Build on 2b’s in-memory run identity and failed-run snapshots by adding explicit queued, running, finalizing, canceled, and failed lifecycle reporting with last-activity data. Preserve truthful partial reports for canceled runs, separate last-attempted from last-successful timestamps, and persist a bounded final report across restarts. Acknowledge Start Sync only after a run is accepted and queued; keep dry runs and legacy terminal status consistent with the actual result. Add paged or incremental details only if measured list size or polling cost warrants it.

## Checklist

- [x] Tests pass locally
- [x] Tests pass on CI
- [x] Changelog is updated


drallgood (changes requested):
## Review: Step 2b — Live per-profile status API

Overall this looks great. Confirmed with `-race`: `internal/sync`, `internal/mismatch`, `internal/api`, and `internal/multiuser` all pass, `go vet` is clean, and `make build` succeeds.

Key things I verified:
- **Auth split is correct.** Per-profile `/status` and `/summary` sit behind `RequireAuth`, and the public `/api/status` projects a scalar-only aggregate so unauthenticated callers never see per-book details.
- **Run isolation is sound.** Generation-based tracking prevents a stale goroutine from overwriting a replacement run's status, and `GetSnapshot` reads under the summary lock, so status/summary always return a coherent snapshot.
- **Outcome accounting reconciles.** `total_books_processed = len(outcome_records)` and the seven mutually-exclusive categories sum to `processed_so_far`; re-recording a book adjusts its old category without inflating totals.
- **Dry-run safety and legacy fields** are preserved, and cancellation correctly clears the snapshot (documented).
- **Mismatch `Collector` refactor** cleanly removes cross-profile shared-state mutation with a compatibility global.

One consistency item I'd like resolved (non-blocking):
- `docs/openapi.yaml:107` documents the profile config field as `dryRun` (camelCase) in the `ProfileConfig` schema, but the actual API field is `dry_run` (snake_case). Every other field — including all the new status fields (`run_id`, `processed_so_far`, `outcome_counts`, `attention_records`, etc.) — is snake_case and matches the database models and web UI. Since Step 3 will build on these field names, it's worth fixing the `dryRun` doc entry to `dry_run` now to settle the convention. (Left as-is, this PR is internally consistent; these new fields should stay snake_case.)

Minor (optional):
- `CancelSync` and `publishFinalStatus` run DB `GetSyncState`/`UpdateSyncState` while holding `syncMutex`, which can stall a concurrent `StartSync`. Consider hoisting the DB I/O out of the lock.
- `applySnapshotToStatus` falls back to showing `processed_so_far` as `books_total` when the pre-count hasn't landed; slightly misleading but documented as legacy behavior.
- Any authenticated session can read any profile's full per-book attention records (no per-profile authz boundary). Pre-existing, but worth keeping in mind for Step 3's profile-ID/route hardening.

**Verdict: Approved** once the OpenAPI `dryRun` → `dry_run` doc fix is in.

Snuffy2:
Thanks for the detailed review. I’ve pushed the requested OpenAPI correction from `dryRun` to `dry_run`.

For the three minor findings:

1. **Database I/O under `syncMutex`:** This is now explicitly included in Step 4’s lifecycle work. It will use per-profile or generation-aware persistence so moving I/O outside the lock cannot allow an older run to publish stale state.

2. **Legacy `books_total` fallback:** This will be addressed in Step 3 while migrating the UI away from the flattened legacy status fields, avoiding an isolated compatibility change before the UI is ready.

3. **Per-profile authorization:** This will also be addressed in Step 3’s profile-ID and route hardening. It needs to be enforced consistently across all profile routes rather than patched only for status and summary.

Thanks again for calling these out.

drallgood (approved):
Thanks — I've verified the `dryRun` → `dry_run` OpenAPI correction on the latest head. That resolves the consistency item fully.

The plan for the remaining three minor findings works for me:
1. Hoisting DB I/O out of `syncMutex` in Step 4, with generation-aware persistence so stale runs can't publish — agreed, and it's the right home since Step 4 adds canceled/failed lifecycle anyway.
2. Deferring the `books_total` fallback cleanup to Step 3 alongside the UI migration — agreed, avoids a throwaway compatibility change.
3. Enforcing per-profile authorization consistently across all profile routes in Step 3 — agreed, that's the correct scope.

**Approved.** Nice work; looking forward to Steps 3 and 4.

PR #185, opened 2026-09-16
Add secure on-demand sync details

Snuffy2:
#### Step 3 of 4 to Improve Sync Status UI

Adds secure, on-demand book-level details for the current sync run while keeping aggregate status polling lightweight. The Sync Status UI now exposes richer outcome details and links, and authenticated profile access follows owner, viewer, administrator, and legacy-ownerless permissions.

- Add a current-run details endpoint with stale-run rejection and redacted aggregate snapshots.
- Enforce profile ownership, viewer read-only access, credential redaction, and non-enumerating authorization failures.
- Show categorized book outcomes with Audiobookshelf, Hardcover, Audible, and Goodreads links where available.
- Harden profile IDs, derived state paths, run identifiers, and async UI session boundaries.
- Document the API, authorization model, unknown-total behavior, and user-visible status workflow.

## Testing Instructions

- `make test`
- `make -B build`
- `make lint` 
- Verify the Sync Status page loads aggregate progress, toggles current-run details, and keeps viewer controls read-only.
- Verify stale or replaced run IDs and unauthorized profile access return `404`.

## Multi-Step Project

~~2b. Live per-profile status API: Publish needs-review, not-found, and failed attention records as attempts finish. Expose a coherent run-scoped snapshot through authenticated per-profile status and summary responses, isolate profiles, and retain legacy response fields so the existing UI remains usable until step 3.~~

3. Sync Status and View Details redesign (this PR): Consume the outcome snapshot, show the full category breakdown, and load details through a secure, run-scoped endpoint. Refresh open details without losing scroll, expanded sections, or focus. Secure profile routes, reject stale responses, clear deleted profiles’ action errors, and remove obsolete UI fallbacks while preserving public API compatibility.

4. Run lifecycle and report history: Add queued, running, finalizing, completed, canceled, and failed lifecycle reporting with last-activity data. Preserve partial canceled and failed reports, separate attempted from successful timestamps, and persist a bounded final report across restarts. Make start acknowledgment, dry-run status, persistence, and cross-profile concurrency generation-safe. Add pagination only if measured detail size warrants it.

## Checklist

- [x] Tests pass locally
- [x] Tests pass on CI
- [x] Changelog is updated


drallgood (changes requested):
Nice work, @Snuffy2. The run-details endpoint (`GET /api/profiles/{id}/runs/{runId}/details`) is well-designed: authenticated, ownership-checked before decryption, `404` for stale/replaced/unknown run IDs with no fall-through to a newer snapshot, and the in-memory snapshot path deliberately avoids hydrating encrypted profile config. The XSS discipline in the frontend is consistently applied (escaping, scheme-validated links, escaped terminal-error text), and I found no data races in the snapshot locking. CI test job passes.

A few things to address before merge:

**1. Token redaction scope is broader than described.** `buildProfileResponseForRequest` (internal/api/handlers_new.go:225) omits `audiobookshelf_token`/`hardcover_token` from *all* profile responses — owners and admins included — but the PR description and commit 6f82fff describe it as viewer-only redaction. Redacting for everyone is a reasonable security posture (and removes the `currentEditUser.audiobookshelf_token` fallback in `handleEditProfile`, app.js:1851, which is now always `undefined`). Please align the docs: update the API description to say tokens are never echoed, note the breaking change for API clients that read tokens back, and drop or comment the stale UI fallback.

**2. Legacy ownerless profiles become admin-only with no claim path.** With auth enabled, `owner_user_id IS NULL` profiles return `404` to non-admins and are filtered from listings; there is no way for a user to claim them. Existing single-user deployments that enable authentication would thereby lose visibility of their own profiles. Keeping admin-only behavior is fine, but please call out this upgrade impact explicitly in README (User Roles section) and MIGRATION.md so users can decide whether to reassign ownership first.

**3. Legacy state-file filename changes for non-unreserved profile IDs.** `profileSpecificStatePath` (internal/multiuser/service.go:797) now appends `encodeProfileID(id)` where develop used the raw ID. For legacy profiles whose IDs contain spaces or other encoded characters, the on-disk file changes name, so incremental sync state is effectively lost and a full re-sync happens after upgrade. Worth a one-time rename fallback (look for the legacy filename first) or at least a migration note.

**4. `WaitForSyncs` is never wired into production shutdown.** It exists (internal/multiuser/service.go:503) and tests use it, but `cmd/audiobookshelf-hardcover-sync/main.go` still returns after `srv.Shutdown` without waiting, so `defer db.Close()` can race in-flight sync goroutines persisting state. One call in the shutdown path closes the gap.

**5. Drive-by: `showTab('profiles')` is a no-op bug.** app.js:1713 calls `showTab('profiles')` after creating a profile, but the only tabs are `users`, `sync`, `add-user`. The selector matches nothing and throws (caught), so the tab never switches and an error toast appears alongside the success toast. Pre-existing, but since this PR rewrote the file it's worth fixing to `showTab('users')` (or whatever the intended tab is).

One note on the lint situation: your local `make lint` block sounds like the cached analyzer/export-data mismatch, but CI runs `golangci-lint@latest` on Go 1.26 and its lint step passes, so no action needed there — worth mentioning so other contributors don't go down that rabbit hole.

Happy to re-review once the docs and the items above are updated.

Snuffy2:
Thanks for the detailed review. I’ve pushed the follow-up changes:

1. **Token redaction:** Updated the OpenAPI description, README, migration guide, and changelog to state that profile APIs never return Audiobookshelf or Hardcover tokens, regardless of authentication state or role. I also removed the obsolete frontend fallback to token fields that are no longer returned. Empty token fields continue to preserve the stored credentials.

2. **Legacy ownerless profiles:** Documented that enabling authentication leaves existing ownerless profiles active for scheduled syncs but makes them visible and accessible only to administrators. There is currently no supported claim or ownership-reassignment workflow; the documented migration path is to create a replacement profile while signed in as the intended owner and then have an administrator remove the old profile.

3. **Legacy state filenames:** Added a one-time migration from legacy raw-ID filenames to the encoded profile-specific path. The migration retains a recoverable `.migrated` backup and avoids unsafe traversal, symlink, and cross-profile path-collision cases. Regression coverage includes spaces, slashes, unsafe path segments, symlinks, an existing canonical state file, and preservation of potential victim files.

4. **Shutdown waiting:** I did not add a standalone `WaitForSyncs()` call after tracing the full scheduler path. The periodic scheduler launches `StartSync` in outer, untracked goroutines before the service wait group is incremented, so the proposed call could return while a queued start still intends to use the database. Active sync contexts are also independent of the shutdown context, making an unconditional wait potentially unbounded. A safe correction needs scheduler admission closure, active-run cancellation, complete goroutine tracking, and a bounded drain together; I’m keeping that work with the planned Step 4 lifecycle changes rather than adding a partial shutdown guarantee here.

5. **Profile creation tab:** Fixed the invalid `showTab('profiles')` call to return to the existing `users` tab after successful profile creation.

drallgood (approved):
Thanks, @Snuffy2 — I've verified each of the follow-ups against the updated branch:

1. **Token redaction** — Confirmed the OpenAPI description, README, MIGRATION.md, and CHANGELOG now state that profile responses never echo tokens regardless of role, and the stale UI fallback (`currentEditUser.audiobookshelf_token` in `handleEditProfile`) is gone. The empty-token-preserves-stored-credentials behavior on the server side (`handlers_new.go`) keeps the edit flow intact, and the removal is behavior-neutral for the UI. One small note: with the fallback gone, external API callers now have *no* way to read tokens back, so the "retain your own values" wording in MIGRATION.md is the right call — a future "rotate token" flow should send the new value explicitly.

2. **Legacy ownerless profiles** — README's User Roles section and MIGRATION.md both document the admin-only behavior and the create-replacement/delete-old workaround clearly. The behavior itself matches what was implemented (repository filters by `owner_user_id`, and `authorizeProfileMetadata` 404s foreign/ownerless profiles to non-admins). Agreed this is a defensible posture for now; if ownership reassignment surfaces as a real need later it can be a tracked feature rather than an upgrade blocker.

3. **Legacy state filenames** — The migration (`migrateLegacyProfileStatePath` in `internal/multiuser/service.go`) is careful: non-symlink regular-file check via `Lstat`, lexical and symlink-resolved containment via `EvalSymlinks`, no-op when a canonical file already exists or when the legacy path equals the canonical path, `.migrated` backup retention, and `safeLegacyProfileIDPathSegments` guarding against `foo/../other` aliasing. The regression tests cover exactly the cases you listed (spaces, slashes, unsafe segments, symlinks, existing canonical state, victim-file non-collision). Note the migration runs inside `StartSync` (`createProfileSpecificConfig`), so the first sync after upgrade performs it — that's the right moment since the state file isn't read elsewhere before then. `make test` all-green with `-race` on `internal/multiuser`, `internal/api`, `internal/sync`, `internal/server`, and `go build ./...` passes.

4. **Shutdown waiting** — Your reasoning holds: `main.go` launches the initial and periodic syncs through untracked `go func(profileID)` wrappers around `StartSync`, so a bare `WaitForSyncs()` call after `srv.Shutdown` could return while a queued admission still increments the wait group, and active run contexts aren't tied to the shutdown context, so an unconditional wait could hang. Acceptable to defer the full lifecycle treatment (admission closure + cancellation + bounded drain) to Step 4 — just keep the existing `WaitGroup` usage scoped to what it actually covers today so the eventual wiring doesn't become misleading.

5. **`showTab('profiles')`** — Fixed to `showTab('users')`; the success-path error toast is gone. Resolves cleanly.

No remaining blockers from my side. Once this is merged and Step 4 lands, we should double-check that migration interplay (legacy path migration happening on first post-upgrade sync) and the shutdown drain against the new lifecycle states.

Approved. 🚀

PR #186, opened 2026-09-17
Add durable sync run lifecycle and history

Snuffy2:
## Summary of Changes

Adds durable per-profile sync run lifecycle and history to Sync Status. Sync starts return the accepted queued run identity, canonical run details survive restarts, and the UI follows each run through queued, running, finalizing, completed, canceled, and failed phases.

- Persist sanitized run reports and retain the newest 10 terminal reports per profile.
- Return the accepted run ID and lifecycle state from sync-start requests.
- Restore completed, canceled, and failed run details after restart.
- Separate last-attempted timestamps from successful non-dry-run sync timestamps.
- Serialize replacement workers, bound shutdown cancellation and draining, and reject stale run updates.
- Remove superseded status, state, mismatch, and test compatibility surfaces.
- Update the web UI, OpenAPI specification, README, migration guidance, Helm values, changelog, and behavioral coverage.

### Coverage retained after legacy-test cleanup

Removing `internal/testutils/incremental_test.go` did not remove incremental-sync coverage: `internal/sync/outcomes_test.go` proves an unchanged book avoids the Hardcover lookup, while `internal/sync/sync_test.go` covers checkpoint persistence, unchanged-book skipping, dry runs, and cancellation. The removed mismatch/prepopulation harnesses are replaced by production-path coverage in `internal/mismatch/mismatch_test.go` (mismatch-to-edition export/input and individual JSON output) and `internal/sync/outcomes_test.go` (sync-owned enrichment/export). The retired `MigrateOldState` v1→v2 test is superseded by `internal/sync/state/state_test.go`, which covers timestamp-only v1 recovery, unversioned and v2 checkpoint preservation/normalization to the current version, and rejection of unknown future versions.

## Multi-Step Project

~~2b. Live per-profile status API: Publish needs-review, not-found, and failed attention records as attempts finish. Expose a coherent run-scoped snapshot through authenticated per-profile status and summary responses, isolate profiles, and retain legacy response fields so the existing UI remains usable until step 3.~~

~~3. Sync Status and View Details redesign: Consume the outcome snapshot, show the full category breakdown, and load details through a secure, run-scoped endpoint. Refresh open details without losing scroll, expanded sections, or focus. Secure profile routes, reject stale responses, clear deleted profiles’ action errors, and remove obsolete UI fallbacks while preserving public API compatibility.~~

4. Run lifecycle and report history (this PR): Track accepted syncs through queued, running, finalizing, completed, canceled, and failed phases with phase-specific and last-activity timestamps. Preserve truthful partial canceled and failed reports, separate attempted from successful non-dry-run timestamps, and retain the newest 10 terminal reports per profile across restarts. Return the authoritative queued run ID at acceptance, expose dry-run state, serialize replacement workers, bound cancellation and shutdown draining, and keep persistence and cross-profile concurrency generation-safe.

## Testing Instructions

- `GOTOOLCHAIN=go1.26.0 make all`
- Start a sync and confirm Sync Status follows its accepted run from queued through a terminal phase.
- Restart the service and confirm retained terminal run details remain available.
- Start, cancel, and replace runs for separate profiles and confirm status and exports remain profile-isolated.

## Checklist

- [x] Tests pass locally
- [x] Tests pass on CI
- [x] Changelog is updated


drallgood (changes requested):
Thanks @Snuffy2 for the masive PR. Quite a bit to review.
The lifecycle design looks correct — per-profile gate serialization, generation-checked stale-report rejection, last_attempted_at/last_successful_at split, sanitized persisted snapshots, and bounded shutdown draining are all well done and well tested.

Some points to check/change:
1. Confirm the deleted coverage is replaced, not lost (main ask). The PR body declares the superseded-surface cleanup, but please add a short paragraph mapping each big deletion to the test that now protects its contract — specifically incremental sync (internal/testutils/incremental_test.go), the mismatch/prepopulation workflow tests, and the v1→v2 state-file migration (MigrateOldState). Incremental behavior is a stated safeguard in AGENTS.md; if any of these deletions has no replacement, restore those tests.

2. Two small robustness nits
- Non-terminal report rows (interrupted "queued"/"running" rows after a crash) are only cleaned by deleteStaleQueuedSyncRunReports on the next accept for that profile (internal/database/repository.go). Consider a startup sweep that marks stale non-terminal rows as failed, so a DB interrupted mid-run doesn't retain a ghost "queued" row for profiles that are never synced again.
- DeleteProfile (internal/multiuser/service.go:193) clears neither latestRuns nor profileGates for the deleted profile. Harmless today because generations are per-profile and DB-derived, but a comment or cleanup would keep the invariants obvious to future readers.

internal/database/repository.go
```diff
@@ -38,6 +40,262 @@ type ProfileWithTokens struct {
 	SyncConfig          SyncConfigData `json:"sync_config"`
 }
 
+const maxSyncRunReports = 10
```
drallgood, line 45: We should make this configurable

internal/sync/state/state.go
```diff
@@ -12,23 +12,12 @@ import (
...
 )
 
-const (
-	CurrentVersion   = "2.0"
-	DefaultStateFile = "./data/sync_state.json"
-)
+const DefaultStateFile = "./data/sync_state.json"
 
 type State struct {
-	Version      string             `json:"version"`
```
drallgood, line 21: Wouldn't it be safer to keep a version? In case we need to update the file in the future?

internal/sync/state/state.go
```diff
@@ -531,22 +453,6 @@ func normalizeProgress(progress float64) float64 {
 	return progress
 }
 
-type v1State struct {
-	LastSyncTimestamp int64  `json:"lastSyncTimestamp"`
-	LastFullSync      int64  `json:"lastFullSync"`
-	Version           string `json:"version"`
-}
-
-func migrateV1ToV2(v1 v1State) *State {
```
drallgood, line 540: Do we migrate from v2 to v3?

internal/sync/service.go
```diff
@@ -1994,8 +2008,8 @@ func (s *Service) processBook(ctx context.Context, book models.AudiobookshelfBoo
 		}
 		s.recordBookOutcomeWithMatchMethod(book, outcomeHint, outcomeReason, outcomeError, hcBook, matchMethod)
 		bookLog.Debug("Book processing outcome recorded", map[string]interface{}{
-			"outcome":               outcomeHint,
-			"total_books_processed": s.processedOutcomeTotal(),
```
drallgood, line 1998: any reason to rename this?

internal/sync/service.go
```diff
@@ -2235,15 +2249,15 @@ func (s *Service) processBook(ctx context.Context, book models.AudiobookshelfBoo
 			// Replace the initial lightweight record with any details already
 			// available from the title/author candidate. The run collector below
 			// remains separate from the live status snapshot.
-			s.enrichLiveMismatch(mismatchData)
```
drallgood, line 2238: I noticed this earlier.. why do we rename mismatches to "attentionCandidates?

web/static/app.js
```diff
@@ -855,9 +869,10 @@ class SyncProfileApp {
     }
 
     async fetchTerminalError(profileId, identity, signal, authGeneration) {
+        const runId = JSON.parse(identity)[1];
```
drallgood, line 872: hmm.. this could fail?

drallgood (changes requested):

Snuffy2:
Thanks for the detailed review.

- Added configurable report retention:
  - `database.sync_run_report_retention` / `DATABASE_SYNC_RUN_REPORT_RETENTION`
  - Defaults to 10, preserves existing behavior, and applies both on writes and at startup when the limit is lowered.

- Restored explicit state-file versioning:
  - New state saves write version `3.0`.
  - Unversioned, v1, and v2 files remain readable.
  - v2 checkpoints normalize to v3 without losing `Books` data; timestamp-only v1 files rebuild checkpoints on the next sync.
  - Unknown future versions fail safely.

- Added startup reconciliation:
  - Interrupted queued/running/finalizing reports are retained as failed with a restart-interruption error.
  - Startup also prunes terminal history to the configured retention limit.

- Fixed profile deletion lifecycle cleanup:
  - Deletion now blocks new admissions, cancels and drains active work, then evicts lifecycle caches.
  - Tombstones prevent late workers from recreating state or persisting reports after deletion.

- Fixed the stale-worker report race:
  - A canceled run replaced by a new run can no longer overwrite its durable canceled report with a late failed result.
  - All terminal persistence now requires the exact current run and generation.

- Confirmed deleted coverage is replaced by production-facing tests:
  - Incremental no-op/checkpoint behavior: `internal/sync` outcome and sync tests.
  - Mismatch/export workflows: `internal/mismatch` and `internal/sync` tests.
  - State compatibility: versioned state migration tests now cover v1, v2, and unversioned files.

- On the naming questions:
  - `attentionCandidates` is transient status metadata, distinct from persisted/exported mismatch artifacts.
  - `processed_so_far` deliberately identifies an in-progress count rather than a final processed total.

- On the UI JSON concern:
  - The parsed identity is generated internally with `JSON.stringify`, and response JSON failures are already caught and retried. No externally reachable failure path was found.

drallgood (changes requested):
Verified all the pushed changes against the updated diff — thanks @Snuffy2 , this addresses the substantive points, including the stale-worker race, deletion tombstones, startup reconciliation, and configurable retention. Happy to see 3.0 state-file versioning restored with tolerant load + future-version rejection.

One last thing on the state files: the versioning is fixed, but the legacy path migration still looks gone — MigrateOldState (./cache/sync_state.json → ./data/sync_state.json) was removed and nothing relocates it now. Impact is low since v1-era files only carried timestamps that now live in the DB, but your reply says v1 files "rebuild checkpoints on the next sync" — that's only true when the file is at the current path. Either keep the one-time relocation, or add a MIGRATION.md note that files in cache/sync_state.json are no longer read (v1-only, superseded by DB timestamps). Once that's stated, I'm good to approve.

drallgood (approved):
Looking good now. Thanks @Snuffy2

PR #188, opened 2026-09-19
fix: preserve ABS finished read date for new books

Snuffy2:
## Summary of Changes

Preserves Audiobookshelf completion dates for newly matched finished books and prevents Hardcover from creating synthetic completion dates when Audiobookshelf has not supplied a finished date.

## What Changed

- Create newly matched completed books as `WANT_TO_READ`, add the Audiobookshelf-dated finished read, then transition them to `FINISHED`.
- Exclude explicitly finished and computed-100% books without an Audiobookshelf `finished_at` from Hardcover matching until a date is available.
- Prevent ownership, edition lookup, mismatch recording, user-book, read, status, and checkpoint work for those undated finished books.
- Preserve safe dry-run behavior and add regression coverage for dated, undated, computed, zero-duration, enhanced-progress, identifier, and dry-run paths.
- Document that existing reads created with the old sync-time synthetic date are not repaired.
- Remove a duplicate user-book creation log entry.

## Why

Creating a Hardcover user book directly as `FINISHED` can cause Hardcover to create a completion dated at sync time. Waiting for a real Audiobookshelf completion date avoids corrupting reading history. Books without that date are skipped before matching so repeated syncs do not perform unnecessary Hardcover lookups or side effects.

## Testing Instructions

- `make test`
- `make build`
- `go test ./internal/sync -race -count=1`

## Checklist

- [x] Tests pass locally
- [x] Tests pass on CI
- [x] Changelog is updated


drallgood (commented):
Thanks for this — the direction is right and it matches the existing "skip missing finished_at" guards already inside `HandleFinishedBook`. I applied the diff on top of current `develop`: `go test ./...`, `go build ./...`, and `go vet ./internal/sync/...` all pass, and CI is green. A few things I'd like to confirm/address before merging:

**1. Computed-finished books without a date are now permanently skipped (biggest behavioral change).**
`isFinishedWithoutFinishedAt` fires not just for `IsFinished=true` but for any book where `determineBookStatus` computes FINISHED via `progress >= 1.0` (i.e. `IsFinished=false`, `CurrentTime >= Duration`, no date). For that case ABS won't necessarily ever populate `finished_at` (the date is only set when the item is explicitly marked finished), so the "remains eligible for a later sync with a date" promise may not hold — these books silently fall out of sync where the old code created a FINISHED user book. Skipping them is consistent (the old synthetic "today" date was the bug), but please confirm this is intentional and note the computed-finished case explicitly in the changelog/README. Also worth stating that since skipped books are deliberately not checkpointed, a permanently-dateless finished book triggers full matching (ASIN/ISBN lookups + `GetEdition`) on every run even under incremental sync.

**2. Existing synthetic dates are not repaired.**
For books that were already created under the old behavior with a sync-time finished read, once the user book is `FINISHED` with a finished read present, `HandleFinishedBook` reports `AlreadyCurrent` and never rewrites `finished_at` — even when ABS later provides a real date. This PR correctly prevents *new* bad dates but doesn't heal *old* ones. That's fine, but it's worth an explicit sentence in the changelog so users know pre-existing reads are left alone.

**3. The skip is broader than "reading state".**
In `processFoundBook`/`findBookInHardcover`, missing-date finished books also skip the `sync_owned` ownership marking, the edition-resolution fallback (`GetEdition` when edition is missing), and the NeedsReview mismatch recording on identifier-matched paths. This is arguably the right conservative choice, but it goes beyond the "reading state unchanged" wording in `processBook` — worth confirming the ownership suppression is intended and capturing it in the changelog entry.

**4. Outcome asymmetry worth noting.**
Undated finished books matched only by title/author still produce a NeedsReview mismatch record (the title/author block runs before the new guard), while identifier-matched ones are silently skipped. Probably fine, just be aware the flows may appear inconsistent in the UI.

**5. Dry-run addendum in `HandleFinishedBook` is effectively dead code.**
In the real `processBook` flow an undated finished book returns at the new guard before `findOrCreateUserBookID` (dry-run sentinel `-1`) or `HandleFinishedBook` are reached, so the new `FinishedAt <= 0 → OutcomeSkipped` branch at `service.go:3052` is defense-in-depth only and the unit test exercises it directly rather than through `processBook`. Suggest adding a dry-run `processBook` test to prove the end-to-end skip (the outcome tests only cover the non-dry-run path), if only to lock in the behavior.

**6. Nit:** two identical `logCtx.Info("Creating new user book with status", …)` statements remain in `findOrCreateUserBookID` (now both carrying `status` + `target_status`); since you touched both lines anyway, drop the duplicate.

Onboarding-wise the new tests are solid (no-mutation assertions, state-not-checked assertions, dated/missing-date/zero-duration paths). Happy to approve once the computed-finished question (1) is answered.

drallgood (changes requested):

Snuffy2:
Thanks for the detailed review. I’ve addressed the points as follows:

1. Undated finished books, both explicitly finished and computed at 100%, are now skipped immediately after progress enhancement, before any Hardcover lookup. They are excluded from matching, ownership, edition resolution, mismatch recording, mutations, and checkpointing until a finished date is available.

2. Existing synthetic reads are intentionally not repaired. I updated both the README and changelog to document this limitation and clarify that the fix applies to newly matched books.

3. The broader side effects are now intentional: because the guard runs before matching, undated finished books receive no Hardcover-related processing at all.

4. This also removes the title/author versus identifier asymmetry for these books; neither path is attempted while `finished_at` is missing.

5. Added end-to-end coverage for the dry-run path, including the skipped outcome, lack of Hardcover calls, and lack of checkpoint state.

6. Removed the duplicate `Creating new user book with status` log statement.

drallgood (commented):
Thanks for the rework — this addresses the substantive points from my previous review. I re-applied the updated branch on top of current `develop` and verified locally: `go test ./...`, `go build ./...`, `go vet ./internal/sync/...`, and `gofmt` are all clean.

Confirmations on the changes:
- Moving the guard to the top of `processBook` (right after progress enhancement) makes the exclusion uniform — no matching, ownership, edition resolution, mismatch recording, mutation, or checkpointing for undated finished books — and removes the title/author-vs-identifier asymmetry. The new `assertNoHardcoverBookSearches` assertions and the `TestProcessBookDryRunFinishedWithoutFinishedAtSkipsHardcoverMatching` end-to-end dry-run test lock that in.
- README/changelog now clearly state the fix applies to new matches and does not repair existing synthetic reads.
- Duplicate `Creating new user book with status` log is gone, and the creation-status conversion is consistently reflected in all log/outcome sites.

Remaining observations (none blocking, mostly maintainability):

1. Several of the new checks are unreachable given the new placement. `finishedWithoutFinishedAt` is computed at the top from `book.Progress`, which is never mutated afterward (`book` is passed by value to `findBookInHardcover`, and enhancement already ran), and the top guard returns whenever it is true. So the second guard at `service.go:2796` and the `&& !finishedWithoutFinishedAt` conditions on the state updates at `service.go:2652`, `2712`, and `2784` can never be true in the main flow. Same applies to the now-redundant inner guards in `findBookInHardcover`/`processFoundBook` and the added dry-run branch in `HandleFinishedBook` (unreachable from `processBook` since the top guard returns first). Keeping a single authoritative gate would avoid confusion about which one is actually doing the work; if you prefer to keep them as defense-in-depth, a brief comment noting they are unreachable today would help reviewers not to re-audit them each time.

2. Because the guard now runs before the test-book filter and the incremental `NeedsSync` check, dateless finished books always surface the `finished book has no Audiobookshelf finished_at` skip outcome rather than "book filter" or "incremental state is current", and are never checkpointed. Fine for production; just noting it for anyone using `TestBookFilter` to iterate on one of these books.

3. Worth double-checking the "computed finished" exclusion once more against real ABS data: a book at 100% progress with `IsFinished=false` and no date is now excluded from matching forever unless ABS later marks it finished (and only then a date appears). The changelog/README already describe this accurately, so this is confirmation of intent rather than a request to change behavior.

The dated happy path is preserved: newly matched finished books are created as `WANT_TO_READ`, then the ABS-dated read is inserted and the status moved to `FINISHED` in the same sync, with `deleteBlankReads` cleaning any auto-created rows. I'm satisfied this is ready.

Snuffy2:
Thanks for the thorough re-review:

1. **Unreachable checks:** Agreed. I went with a single authoritative gate rather than defense-in-depth. The guard at the top of `processBook` is now the only check for undated finished books. I removed:
   - the second guard and the three `&& !finishedWithoutFinishedAt` conditions on the state updates in `processBook`
   - the inner guards in `findBookInHardcover` and `processFoundBook`, which now match `develop` again
   - the dry-run branch in `HandleFinishedBook`, along with the test that only covered that branch

   I added a comment on the top guard saying it is the single gate and that nothing downstream re-checks. The end-to-end `processBook` tests, including the `assertNoHardcoverBookSearches` assertions, still cover the behavior. `gofmt`, `go build ./...`, `go vet ./internal/sync/...` and `go test ./...` all pass.

2. **Guard runs before the test-book filter and incremental check:** Acknowledged. This ordering is intentional so the skip is uniform and undated finished books are never checkpointed. I've left it as is. Anyone using `TestBookFilter` on one of these books will see the `finished_at` skip reason instead of "book filter".

3. **Computed-finished exclusion:** Confirmed this is the intended behavior. A book at 100% progress with `IsFinished=false` and no date is excluded from matching until ABS supplies a finished date. The README and changelog already describe this, so I made no change.

drallgood (approved):
Verified the finalized version on a fresh `origin/develop` base: the diff is now limited to the single authoritative gate in `processBook` (with the explicit "single gate, nothing downstream re-checks" comment), the `WANT_TO_READ` creation-status conversion, the `isFinishedWithoutFinishedAt` helper, and the e2e tests in `outcomes_test.go`. No dead state gates, no redundant inner guards, no `HandleFinishedBook` dry-run branch.

Local checks on top of `develop` all pass: `go test ./...`, `go build ./...`, `go vet ./internal/sync/...`, and `gofmt`. CI is green on the head commit.

Outstanding points are fully resolved:
- The single-gate refactor is clean — I confirmed `book.Progress` is not mutated after the gate (`enhanceBookProgressFromUserData` runs before it, and `findBookInHardcover` takes the book by value), so no downstream re-check is needed.
- `TestBookFilter`/incremental-skip ordering is documented as intentional (uniform skip, never checkpointed).
- Computed-finished exclusion is confirmed intended and documented in README/changelog.

The dated happy path remains intact: new matches are created as `WANT_TO_READ`, the ABS-dated read is inserted, and the status is moved to `FINISHED` in the same sync, with `deleteBlankReads` as the cleanup for auto-created rows. Approving.

PR #190, opened 2026-09-20
fix: detect ebooks and match them to Hardcover ebook editions

Snuffy2:
## Summary of Changes

Audiobookshelf reports `mediaType: "book"` for ebooks as well as audiobooks, so every check for `mediaType == "ebook"` never fired. As a result, Sync Status labeled ebooks "Audiobook", `include_ebooks: false` did not skip them, and they were matched only against Hardcover audiobook editions, which left them in needs-review or not-found.

- Add `AudiobookshelfBook.IsEbook()`, which decides from the media content: an ebook file or format and no audio. Audio is detected the same way Audiobookshelf itself does (its `hasAudioTracks` rule): `numTracks` (non-excluded audio files) or a duration, so audio files marked excluded do not count. `numTracks`, `ebookFile` and `ebookFormat` are ordinary fields on the media struct, so each library item is still decoded once and `IsEbook()` works for items built in code as well as decoded ones. An audiobook that also carries an ebook file stays an audiobook. The legacy `mediaType == "ebook"` value is still honored.
- Use it for the Sync Status format label, the `include_ebooks` skip in `processBook`, and the Hardcover reading format put on the per-book context. ASIN, ISBN and title/author lookups now resolve ebook editions (reading format 4). The mismatch collector builds its own context, so the format is also passed to it through a new `MediaMetadata.ReadingFormat` field and applied to its enrichment lookups.
- Share one `readingFormatIDFromCtx` helper in the Hardcover client, and restrict `GetBookByID` to editions of the requested reading format (a book with no edition of that format gets none, instead of falling back to another format).
- Update the config comments, README and CHANGELOG to say "ebook-only".

**Behavior change:** with the default `include_ebooks: false`, ebook-only items are now skipped, as the option is documented. Previously they were synced against audiobook editions, so users who relied on that should enable `include_ebooks`.

**Not addressed:** in-progress ebook progress (progress is derived from audio duration, so it reads 0%; finished ebooks sync).

## Testing Instructions

- `go test ./...` and `node --test web/app.test.js` pass.
- New tests: `internal/models/audiobookshelf_test.go` (IsEbook on expanded, minified, excluded-audio and legacy payloads), `TestOutcomeRecordFormatReflectsMediaContent`, `TestProcessBookSkipsAudiobookshelfEbooksUnlessIncluded` (fails without the fix), and `internal/api/hardcover/reading_format_test.go` (checks the format ID sent to the GraphQL API), and `internal/mismatch/reading_format_test.go` (an ebook-sourced mismatch enriches against ebook editions; fails without the fix).

## Checklist

- [x] Tests pass locally
- [x] Tests pass on CI
- [x] Changelog is updated


drallgood (changes requested):
## Summary

This is a well-diagnosed and well-tested fix. The root-cause analysis matches the
Audiobookshelf behavior I'd expect (mediaType "book" for both formats), the model
change is self-contained, the tests are observability-focused and include a
"fails without the fix" case, and the Shared-helpers refactor (readingFormatIDFromCtx)
removes real duplication. Docs/config/CHANGELOG are updated consistently, including
honest disclosure of the behavior change.

A few things to consider before merge:

### 1. Mismatch enrichment does not actually carry the reading format (functional gap + description mismatch)

The PR description states "mismatch-enrichment lookups now resolve ebook editions
(reading format 4)", but `Collector.AddWithMetadata`
(`internal/mismatch/mismatch.go:76`) builds its own context from
`context.Background()` and receives no reading format through
`addMismatchWithMetadata` (`internal/sync/service.go:724`). So the GetBookByID /
SearchBookByASIN / SearchBookByISBN enrichment calls inside it always default to
format 2 (audiobook). Combined with the new "no fallback" rule in GetBookByID,
an ebook-only Hardcover book in the enrichment path now returns **no edition**
where the old code would at least have fallen back to the ebook edition and
populated identifiers (ASIN/ISBN/publisher). For ebooks this makes the enrichment
strictly less useful than before the PR.

Please either thread the reading format into the mismatch collector (e.g., a
parameter on AddWithMetadata or a field on MediaMetadata/BookMismatch), or
soften the PR description / add a follow-up. It would be good to add a test that
an ebook-sourced mismatch enriches against an ebook edition.

### 2. `hasEbook` detection via stringified RawMessage is brittle

`string(raw.Media.EbookFile) != "" && string(raw.Media.EbookFile) != "null"`
treats an empty object `{}` as an ebook and is fragile to representation changes.
A presence check (RawMessage nillable, or checking whether the key exists) would
be more robust than string comparisons against the literal `"null"`.

### 3. Double decode cost + payload-shape risk in UnmarshalJSON

Every library item is now parsed twice (`json.Unmarshal` into the struct plus the
raw media struct) — worth a comment for large libraries. More importantly, the
`tracks: []json.RawMessage` field only matches payload shapes where `tracks` is an
array. The client requests `minified=0` so this holds today, but a non-expanded
response (or a future ABS version) that reports `tracks` as a count would fail the
whole per-book decode and abort the library fetch. Consider a defensively tolerant
decode (e.g., handle both) or a comment documenting the assumption, and keep an
eye on the test payloads matching what your ABS version actually sends.

### 4. Minor nits

- `IsEbook`'s `Duration <= 0` is redundant with `hasAudio` for everything decoded
  via `UnmarshalJSON`, but is what saves objects built without it; a comment
  clarifying that `IsEbook` depends on the UnmarshalJSON parse path would help.
- The `content` field has no `json:"-"` / omitempty tag, so debug dumps
  (`first_book_json`) will now carry the synthesized `content` object — fine for
  debugging, but confirm that's intended.
- The default-config behavior change (ebook-only items now skipped instead of
  synced to audiobook editions) is disclosed and documented; consider a
  one-line "breaking for users who relied on the old behavior" note in the
  changelog entry.

### Verdict

Logic is sound, tests cover the observable contracts, and the intentional
behavior change is clearly documented. Approve once the mismatch-enrichment gap
(#1) is either fixed or explicitly spun into a tracked follow-up, and #2/#3 are
acknowledged.

Snuffy2:
Thanks for the careful review. 

**1. Mismatch enrichment and the reading format.** You were right, and my description overstated it. `AddWithMetadata` builds its own context, so the enrichment lookups defaulted to audiobook. Combined with the strict `GetBookByID`, ebooks would have lost enrichment. I added a `ReadingFormat` field to `MediaMetadata`, set it at all five call sites in the sync service, and apply it to the collector's enrichment context. `TestAddWithMetadataEnrichesAgainstTheSourceReadingFormat` checks the format ID sent to Hardcover for ebook, audiobook and the default, and it fails for ebooks without the change.

**2. `hasEbook` via a stringified `RawMessage`.** `ebookFile` is now a `*json.RawMessage`, so a JSON null decodes to nil and presence is a nil check, with no string comparison. An empty `{}` still counts as an ebook, which matches Audiobookshelf's own `!!ebookFile` truthiness. There's a test for that case.

**3. Double decode and payload shape.** I removed the double decode entirely. `numTracks`, `ebookFile` and `ebookFormat` are now regular fields on the media struct, so there's no custom `UnmarshalJSON` and each item is parsed once. I also dropped the array-typed `tracks` field, so an unexpected shape there can't fail the whole library fetch. Audio detection now uses `numTracks` (present in both minified and expanded items) plus `duration`, following Audiobookshelf's own `hasAudioTracks`, which ignores excluded audio files. I confirmed those shapes from the Audiobookshelf source.

**4. Nits.**
- `IsEbook` no longer depends on the decode path, so items built in code behave the same as decoded ones, and the `Duration` guard question no longer applies.
- On `content` appearing in `first_book_json`: that field no longer exists. It was unexported, so `encoding/json` would not have serialized it anyway. The new fields are ordinary exported ones and will appear in that debug dump, which is harmless.
- I added the behavior-change note to the changelog and to the PR description: ebook-only items are now skipped by default, and users who relied on the old behavior should enable `include_ebooks`.

drallgood (approved):
Re-reviewed the updated commits (5e0a20b, f5960d6): all four points are addressed and verified in the code.

1. Mismatch enrichment now resolves the source item's format: `ReadingFormat` on
   `MediaMetadata`, set at all five `MediaMetadata` sites in `processBook`, applied to
   the collector's Hardcover enrichment context, defaulting to audiobook when empty.
   `TestAddWithMetadataEnrichesAgainstTheSourceReadingFormat` covers ebook/audiobook
   and the default at the GraphQL boundary.
2. `hasEbook` is a nil-check on a `*json.RawMessage` field (JSON null → nil), with
   ABS-parity behavior for an empty `{}` object and tests for both.
3. The second JSON parse is gone; the audio/ebook fields are regular exported media
   fields, the array-typed `tracks` was dropped, and the "tracks as a count" shape is
   regression-tested.
4. `IsEbook` is now a pure function of the exported fields plus the legacy media type,
   with a dedicated non-decoded-path test, and the changelog leads with the behavior
   change and `include_ebooks` guidance.

Approve. No remaining blocking items.

One request for the future:
Could you please not force push? Makes re-reviews a bit challening

PR #191, opened 2026-09-20
fix: move logging outside mutexes

Snuffy2:
## Summary of Changes

Moves log writes out of mutex-protected sections so a slow or blocked log writer can no longer hold a lock that other goroutines need, and makes the current-user lookup safe against a panicking request. Log messages, fields, and levels are unchanged.

- Rate limiter: log entries are collected while `mu` is held and written after it is released, covering admission, backoff, reset, and rate-limit header processing.
- Hardcover client: `GetCurrentUserID` uses a single in-flight fetch that waiters can leave on context cancellation, instead of holding the write lock across the API call. The in-flight marker is released from a deferred call, so a panic in the request path cannot leave later callers blocked.
- `state.UpdateBook`, `mismatch.Collector.Add`, and `Service.logSyncSummary` unlock before logging.
- OIDC: `cleanupExpiredStates` returns the expired count and `GetAuthURL` logs it after releasing `statesMutex`.

## Testing Instructions

- `make test` passes locally (race detector, atomic coverage).
- New tests block the log writer and assert the rate limiter's mutex stays free (`Acquire` backoff logging and pacing-recovery logging, independent of test order), cover concurrent `GetCurrentUserID` callers, and cover recovery after a panicking fetch.
- Run the new tests in isolation to confirm they do not depend on test order: `go test -race -count=1 ./internal/util -run 'LogsOutsideLock'` and `go test -race -count=1 ./internal/api/hardcover -run 'GetCurrentUserID'`.
- No user-facing configuration or behavior changes, so no README or documentation updates were needed.

## Checklist

- [x] Tests pass locally
- [x] Tests pass on CI
- [x] Changelog is updated


drallgood (changes requested):
## Review: move logging outside mutexes

**Overall: approving direction; this is a well-scoped, carefully-executed change.** I read the diff, the full head-tree of every touched file, and confirmed against `develop`. The headline guarantees hold:

- All log writes in the rate limiter, `GetCurrentUserID`, `UpdateBook`, `Collector.Add`, the sync summary/ASIN cache, and OIDC cleanup now run with the relevant mutex released.
- Message text, fields, and levels are verbatim from the base (I checked each call site).
- The `GetCurrentUserID` defer/`finishCurrentUserFetch` design is sound: marker creation/teardown are both under the mutex, the channel is closed outside the lock, and there's no lost-wakeup or double-close path. Panic-clears-marker is a nice catch.
- The new tests are genuinely order-independent (global `zerolog` level is guarded by `t.Cleanup`) and the blocking-writer probe directly exercises the "mutex stays free while logging" contract.

A few observations, none blocking:

1. **Rate-limiter permit leak under panic (out of scope, but adjacent).** `executeGraphQLOperation` calls `release()` only on the normal error paths, not via a deferred/guaranteed path. If `httpClient.Do` panics (which is exactly the scenario this PR makes the client resilient to), a semaphore permit is lost. After `DefaultMaxConcurrent` such panics, `Acquire` would block forever. It's pre-existing, but since this PR is the one asserting panic-resilience, it's worth adding a `defer`-style release and a panic test at `maxConcurrent = 1` — the current panic test passes even if the permit leaks because the limiter is constructed with 2 slots.

2. **`applyExponentialBackoff` discards the `setRate` return value.** Harmless today (an exponential backoff always grows the rate, so "pacing recovered" can't fire there), but the asymmetry is undocumented — a one-line comment, or threading the entry up, would prevent a future footgun.

3. **Minor:** on a failed fetch, all waiters wake and re-enter the loop, so a sustained API failure causes one fresh fetch per waiter (matching pre-PR burst behavior). A brief comment in the loop explaining the error-path semantics would help maintainers.

Thanks for the careful work here — particularly the header-processing refactor and the test discipline.

Snuffy2:
Thanks for the careful review.

1. **Permit leak under panic.** Fixed. `executeGraphQLOperation` now releases the permit through a once-guarded deferred call, so a panic in `httpClient.Do`, the body read, or header processing can't leak a concurrency slot. The explicit releases still free it early on the normal paths. I agree it predates this PR (the same three `release()` calls are on `develop`), but it undercut the panic-resilience claim. I also changed `TestGetCurrentUserIDRecoversAfterFetchPanic` to use a 1-permit limiter, so a leaked permit now makes the follow-up call fail with `rate limiter error: context deadline exceeded`. I confirmed the test fails with the deferred release removed and passes with it.

2. **`applyExponentialBackoff` discarding `setRate`'s return value.** Added a comment saying a backoff never lowers the rate, so `setRate` can't return its "pacing recovered" entry there, and that it should be handled if `exponentialBackoff` can ever return a lower rate. I left the signature alone rather than thread an entry that is always nil.

3. **Waiter semantics after a failed fetch.** Added a comment in the `GetCurrentUserID` wait loop. After a failed fetch each waiter re-checks the cache and may start its own fetch, which matches the pre-PR behaviour of retrying a failed lookup per caller.

drallgood (approved):
Verified the follow-up commit cce0d26c against my three comments: the once-guarded deferred permit release in executeGraphQLOperation is correct (including the per-attempt sync.Once, which prevents both double-release and over-release into the semaphore), the 1-permit panic regression test now genuinely fails on a leaked permit, and both new comments accurately document the backoff/setRate and failed-fetch waiter behavior. Changelog is consistent and CI is green. Ready to merge.

Thanks @Snuffy2

PR #192, opened 2026-09-20
fix: suppress misleading request logs during daily quota pause

Snuffy2:
## Summary of Changes

While an exhausted Hardcover daily quota is holding requests until its reset, the sync no longer logs "about to send this request" messages for requests that are not actually being sent. Once the quota resets, logging returns to normal. Pacing and admission behavior are unchanged.

- Added `RateLimiter.DailyQuotaPaused()` (and `Client.DailyQuotaPaused()`), backed by a logging-only `dailyPauseUntil` marker set when an authoritative daily reset wait is applied and cleared by `ResetRate()`. Admission still uses `backoffUntil`.
- The Hardcover client now acquires its rate-limiter permit before logging `Executing GraphQL request` / `GraphQL request body`, so those logs describe an admitted request rather than one still waiting.
- Request-intent logs in the Hardcover client (ISBN/book/person searches, person details, GraphQL query) and the sync service's search-intent logs (ASIN, ISBN, title/author, read-status fetch) are skipped while the daily pause is active. The sync service checks this through an optional `DailyQuotaPaused()` interface, so test and alternate clients are unaffected.

Why: with the daily quota exhausted, every queued request logged its intent and then sat waiting for the reset. The logs read as if requests were being made and failing, which hid the single warning that explains the pause.

## Testing Instructions

- `make test` (race detector) passes locally, and `make build` succeeds. `go vet ./...` and `gofmt` are clean. 
- New tests cover the limiter pause state and its logs (`internal/util`), GraphQL request logs while waiting versus once admitted (`internal/api/hardcover`), and sync search-intent logs during and outside a pause (`internal/sync`).

## Checklist

- [x] Tests pass locally
- [x] Tests pass on CI
- [x] Changelog is updated


drallgood (changes requested):
## Review

Nice, focused change. The core idea — a logging-only `dailyPauseUntil` marker alongside the existing `backoffUntil` admission pause — keeps risk low, and the tests are genuinely behavior-focused rather than structure-bound. I re-read these against the current limiter state machine and ran the new tests locally on a clean checkout; they pass, and CI is green/mergeable.

### What works well
- Moving `Acquire()` above the `"Executing GraphQL request"`/`"GraphQL request body"` logs in `executeGraphQLOperation` is the cleanest part of the fix: logs now describe an admitted request by construction, with no per-site gating needed on the hottest path.
- The sync-service optional-interface check (`s.hardcover.(interface{ DailyQuotaPaused() bool })`) avoids touching `MockHardcoverClient` and alternate clients. Neat.
- The `setRate` guard is subtle and correct: on the first exhausted response the rate can drop back to `minRate`, which previously would have logged a misleading "pacing recovered" while a pause was being installed.

### Questions / suggestions

1. **The marker and `backoffUntil` can diverge; logs can return while requests are still held.** They start aligned, but a subsequent `applyExponentialBackoff` or Retry-After longer than the daily reset pushes `backoffUntil` past `dailyPauseUntil` (e.g., daily reset 5 min, then an unrelated Retry-After 10 min). Between 5 and 10 minutes `DailyQuotaPaused()` returns false while admission is still blocked — the exact misleading-log window this PR is meant to close. Also, `applyDailyResetWait` sets `dailyPauseUntil = now + delay` even when `applyPause` does *not* extend `backoffUntil` (an existing longer backoff), so the marker is better read as "the most recent daily wait attempt" than "the daily pause." Worth a comment documenting this, or gating on whether the daily wait actually extended `backoffUntil`.

2. **The guards added inside `applyLegacyHeaders` only matter in a mixed-format environment.** Since the marker is set solely by an exhausted IETF `daily` bucket, and legacy exhaust installs its own `applyRetryAfter` pause without setting the marker, the new legacy-header suppression ("Approaching rate limit…", "Rate limit will reset…") is inert unless an IETF daily pause was established earlier and a legacy-format response arrives during it. If legacy-header exhaust is out of scope (which seems reasonable), consider dropping those two guards to keep the change honest about what it covers — or add a one-line comment when they're added that they only apply when a prior IETF daily pause is active.

3. **Call-site-by-call-site gating is incomplete and easy to regress.** Pre-admission intent logs that are not gated will still spam during a pause, e.g. `UpdateUserBookRead`'s "Updating user book read" (client.go:2328), "Getting edition details for user book creation" (client.go:3365), and service-side "Updating existing read status to mark as finished" / "Updating book status to FINISHED" (service.go:3222, 3316). The low-level path is handled by the Acquire reorder, but anything logged before `executeGraphQLOperation` still prints while held. Since the whole point of the PR is that these logs are misleading during a pause, a single request-intent logging helper that all pre-admission intent logs go through would make the guarantee structural rather than a matter of remembering each new call site. At minimum, worth a comment at the log sites noting that pre-admission intent logs must be gated.

4. **Behavioral side effect of the reorder**: a request that fails `Acquire` (context cancellation/deadline) no longer logs `"Executing GraphQL request"` at all, since the log moved after the acquire. That matches the intended suppression, but it slightly reduces visibility into limiter-bound stalls — the only signal left is the warn "Daily rate limit exhausted, pausing until reset" during an active pause. Fine by me, just confirming that's intended.

5. **Housekeeping**: the branch still contains commit `6eae042` (the rate-limiter rework), which is already merged into `develop` via #176 — rebasing onto `develop` would drop it and show only the two real commits. Also, the Checklist has "Tests pass on CI" / "Code quality checks pass on CI" unchecked, but CI is green and mergeable now, so those can be ticked (or the run linked).

### Minor / test notes
- `TestRateLimiterAdaptivePacingLogLevels` isn't really about the daily pause — it pins levels of two unchanged log paths. Fine as regression documentation, but it reads as scope creep; the "adaptive pacing" name is also stale post-#176 (burst/token-bucket terminology was removed).
- The new tests call `zerolog.SetGlobalLevel` with `t.Cleanup` restore. testify's default parallelism is 16 (`-test.parallel 16`), so global log-level mutation can race with unrelated tests in the same binary. This pattern pre-exists (client_graphql_test), but each new use adds exposure — if flakes ever appear, this is the first suspect.
- `TestGraphQLRequestLogsWaitForAdmissionDuringDailyPause` is well-constructed: because the logs only emit after `Acquire` returns, the 25 ms-deadline assertion can't flake from a slow machine, and `requests == 0` proves the request never left. Nice.

### Verdict
No blockers from my side — the change is correct for the IETF-daily path it targets, well tested, and CI is green. Items 1–3 are worth resolving or explicitly documenting before merge; the rest is polish.

Snuffy2:
Thanks for the thorough review. I addressed everything without force-pushing, as a normal commit, a merge of `develop`, and two test commits.

1. **Marker and `backoffUntil` diverging: fixed.** `dailyPauseUntil` now tracks the actual admission hold. `applyDailyResetWait` takes the current `backoffUntil` when it is later than the daily reset, and `setBackoffUntil` extends the marker while a daily pause is active. So a longer Retry-After or backoff keeps logs suppressed for as long as requests are held. I added `TestRateLimiterDailyPauseLastsAsLongAsRequestsAreHeld`, which starts a 1s daily reset, adds a 3s Retry-After, and checks the pause is still reported after the daily reset time. I also updated the field and method comments to say what the marker means.

2. **Legacy-header guards: dropped.** You're right that they only mattered after an earlier IETF daily pause, and no test covered them. The two legacy logs ("Approaching rate limit…" and "Rate limit will reset…") behave as they did before this PR. The IETF "window nearly exhausted" and "pacing recovered" guards stay, reapplied to #191's collect-then-write log entries.

3. **Ungated intent logs: fixed structurally.** Each package now has one `debugRequestIntent` helper, with a comment that pre-admission intent logs must go through it. All the existing gated sites use it, and so do the four you found: "Updating user book read", "Getting edition details for user book creation", "Updating existing read status to mark as finished" and "Updating book status to FINISHED". Post-admission logs need no gating.

4. **Failed `Acquire` no longer logs "Executing GraphQL request": intended.** Those logs now describe admitted requests, and the "Daily rate limit exhausted, pausing until reset" warning covers the stall.

5. **Housekeeping.** Fixed

**Test notes:**
- `TestRateLimiterAdaptivePacingLogLevels` was scope creep, so I removed it.
- For `SetGlobalLevel`, I added `testutils.SetGlobalLogLevel(t, level)`. It holds a mutex until the test ends and restores the previous level, so overlapping tests can't leave the level wrong. I used it in the three tests this branch added, and also in the two existing tests from #191 that had their own set-and-restore (admittedly also scope creep). There is no `t.Parallel()` in these packages today, so this is defensive.

drallgood (approved):
All review items are resolved — I re-tested the updated branch locally on a clean
checkout (applied `c4cf595` diff onto `4eb2a89`): `internal/util`,
`internal/api/hardcover`, and `internal/sync` all pass, including the new
`TestRateLimiterDailyPauseLastsAsLongAsRequestsAreHeld`. CI is green on both
matrix runs.

1. **Marker/`backoffUntil` divergence: verified fixed.** `applyDailyResetWait`
   adopting the later `backoffUntil` and `setBackoffUntil` extending the marker
   while a daily pause is active closes both divergence directions; the marker
   now tracks the actual hold, and the new test covers the over-reset Retry-After
   case. The updated field/method comments now describe intent accurately.
2. **Legacy guards: dropped as agreed.** The two legacy logs behave exactly as
   before the PR.
3. **Structural gating: verified.** `debugRequestIntent` in both the client and
   the sync service, with the "must go through this helper" contract documented,
   and all 10 pre-admission sites (including the four previously ungated)
   converted. This makes the guarantee durable rather than per-call-site.
4. **Acquire-failure log silence: intended, fine.**
5. **Housekeeping: confirmed** — adaptive-pacing test removed;
   `testutils.SetGlobalLogLevel` is a neat serialization/restore pattern and the
   retrofit into the two #191 tests is reasonable hygiene.

No remaining blockers from my side. One minor, non-blocking note:
`SetGlobalLogLevel` holds the mutex until `t.Cleanup` runs; if a future test
using it ever hard-crashes (vs. fails), the mutex would be left locked and
subsequent log-level tests would hang. Worth a one-line comment or awareness,
but the same applies to many testify patterns.

This is a solid fix — happy to approve.

PR #195, opened 2026-09-21
feat: ISBN and reading-format foundations for mismatch exports

Snuffy2:
## Summary of Changes

This is the first step of a multi-step project that lets a needs-review book get a new Hardcover edition created from its Audiobookshelf item. It adds the shared building blocks the later steps rely on and fixes four problems in the mismatch export along the way. Ordinary syncing is unchanged. Only the mismatch JSON export (`edition_*.json`) changes: for ebook items, for unresolved publishers, for hyphenated ISBNs, and for audiobooks that Audiobookshelf marks abridged.

- Add an `internal/isbn` package that normalizes and parses ISBN-10 and ISBN-13 values, tolerates hyphens and spaces, and converts between the two forms with checksum validation. `isbn.Result.ISBN10()`, `ISBN13()` and the counterpart accessors are used by the edition creator in the next step, so they have no production caller in this PR.
- Add shared reading-format helpers in `internal/models` (`AudiobookshelfBook.ReadingFormat`, `ReadingFormatID`, and the reading-format context helpers). The Hardcover client and the sync service now use them instead of their own local copies.
- Keep hyphenated ISBNs in the mismatch export. ISBN-10 and ISBN-13 were told apart by exact string length, so an ISBN written with hyphens or spaces was dropped. It is now normalized first, and only the form the item carries is set. An ISBN with a wrong check digit is still exported as given, and the export adds `isbn_10_valid` / `isbn_13_valid` (omitted when that ISBN is empty) to report whether the check digit is correct, like Hardcover's fields of the same name. The `edition` command ignores them, so a later step can decide what to do with an invalid ISBN without changing this export.
- Export an unresolved publisher as `publisher_id: 0` instead of defaulting to `1`, and export a publisher that was resolved during export, which was previously dropped.
- Export ebook items as ebook editions: an `Ebook` edition format regardless of any label the record carries, `reading_format: "ebook"`, no `Unabridged` default, `audio_seconds: 0`, and no edition information unless the record carries a real value (placeholders and debug text are dropped), including when the Hardcover identifier lookup failed.
- Export an audiobook that Audiobookshelf marks abridged as `edition_information: "Abridged"`. The `abridged` flag was not read before, so every audiobook exported `Unabridged`. This changes existing audiobook export output for those books only; every other audiobook still exports `Unabridged`.
- Remove the unused `mismatch.ToEditionInput` and `EditionCreatorInput`; the edition import tool consumes `EditionExport`.

## Multi-Step Project

1. ISBN and export foundations (this PR): Add shared ISBN normalization and reading-format helpers, and fix the mismatch export (hyphenated ISBNs are kept, no default publisher, ebook items export as ebook editions, an abridged audiobook exports as Abridged).

2. Edition creator hardening: Make the edition creator reuse an existing edition of the same book by ASIN or ISBN and refuse another book's, honor the requested edition format, send the Audiobookshelf token only to its own server, report cover failures, and create ebook editions.

3. Edition draft endpoint: Add a read-only endpoint that drafts a new Hardcover edition from a needs-review book's Audiobookshelf item, with author, narrator, and publisher resolved.

4. Edition create endpoint: Add the endpoint that creates the drafted edition on Hardcover, with request validation, a per-book in-flight guard, shutdown draining, and dry-run support.

5. Sync identifier matching: Match an existing edition by ASIN, ISBN-13, or ISBN-10 (either form) during sync so the book no longer lands in Needs review.

6. Immediate read-status resync: After an edition is created, re-sync that one book's read status right away, without overlapping a full sync and without waiting for the next one.

7. Sync Status UI: Add an "Add edition to Hardcover" action on needs-review records, with a preview, confirmation, and a resync option.

[Full Plan Document](https://github.com/Snuffy2/audiobookshelf-hardcover-sync/blob/docs/needs-review-edition-plan/docs/implementations/needs-review-edition-creation.md)

## Testing Instructions

From the repository root:

```sh
gofmt -l .        # prints nothing
go build ./...
go vet ./...
make test
node --test web/app.test.js
make lint         # CI pins Go 1.26; run it with a Go 1.26 toolchain
```

The behavior is covered by these tests:

- `internal/isbn`: `TestNormalize`, `TestParse`, `TestResultForms`, `TestSplit`. `TestParse` also pins `Valid`.
- `internal/models`: `TestReadingFormatID`, `TestAudiobookshelfBookReadingFormat`, `TestReadingFormatContext`, `TestAudiobookshelfMetadataDecodesAbridged`.
- `internal/mismatch`:
  - ISBNs: `TestAddWithMetadata_ISBNForms` (hyphenated, spaced, dotted, lowercase `x`, 979, bad checksum and wrong-length values go to the right field with no derived counterpart).
  - ISBN checksum flags: `TestToEditionExport_ISBNChecksumFlags` (valid, 979, and bad-checksum ISBNs keep their value and get the right flag; an empty or non-ISBN value has no flag).
  - Publisher: `TestAddWithMetadata_PublisherID` and `TestToEditionExport_PublisherResolvedDuringExport` (publisher `0` when unresolved, a resolved publisher exported).
  - Ebook export: `TestEbookMismatchExportsAnEbookEdition`, `TestEbookExportIgnoresAnAudiobookFormatLabel`, `TestSavedExportKeepsTheCanonicalReadingFormatAndNoPlatformInfo`, `TestExportEditionInformation`.
  - Abridged: `TestAddWithMetadataCarriesAbridgedIntoTheExport`.
  - Audiobook export unchanged: `TestAudiobookExportKeepsItsFields` (edition format labels, rounded audio length, `Unabridged`, language and country, cover order) and `TestAddWithMetadata_ReleaseDate` (date normalization and the year-only fallback).
- `internal/sync`: `TestProcessBookIdentifierFailureMismatchExportsByReadingFormat` (a record built after a failed identifier lookup exports as an ebook, or as an abridged audiobook, with the right ISBN field).

## Checklist

- [x] Tests pass locally
- [x] Tests pass on CI
- [x] Changelog is updated


drallgood (changes requested):
## Review: ISBN and reading-format foundations for mismatch exports

Reviewed the full diff (14 files, +851/−320) against `develop`, including the
callers in `internal/sync/service.go`, the importer in `internal/edition/`,
and the `cmd/edition` tool. This is a well-structured, well-tested
foundation PR — the refactors are behavior-preserving, the four export
fixes are real bugs, and each has a focused test. Two points below I'd like
to settle before merge; the rest are nits.

### Verified
- Moving the reading-format context key from `hardcover` (string key
  `"hardcover_reading_format"`) to `models` (struct key) is safe: no other
  code reads or writes the old key.
- `publisher_id: 0` is safe for the current importer: `internal/edition/
  creator.go` only emits `publisher_id` when `> 0`, so 0 is omitted from the
  mutation. The fix only changes behavior for direct JSON consumers (the
  future create endpoint), as the changelog says.
- Removing the `export.EditionInfo = "Audiobookshelf"` fallback in
  `saveToFile` is safe: the audiobook branch always produces at least
  `"Unabridged"`, so the fallback could only fire for ebooks.
- Removing `ToEditionInput` / `mismatch.EditionCreatorInput` is safe (`cmd/
  edition`'s `EditionCreatorInput` is an alias of `edition.EditionInput`,
  unrelated).
- `metadata.abridged` is a real ABS field (`media.metadata.abridged`, set
  from `mediaExpanded.abridged` server-side), so decoding it is correct.
- The audiobook export contract is otherwise unchanged: label rules,
  rounded duration, language/country constants, cover preference, and
  "real value wins" for `edition_information` are all pinned by tests.

### Question 1 — checksum-invalid ISBNs (wants a decision)
`isbn.Parse`/`isbn.Split` accept shape-valid ISBNs **regardless of
checksum**, and `TestAddWithMetadata_ISBNForms` intentionally pins that a
bad-checksum value is still exported into `isbn_10`/`isbn_13`. I'm not sure
that's the right call, and this PR is the moment to decide because later
steps build on these fields (step 2 matches/creates editions by ISBN; step 5
matches an existing edition by ISBN-13 or ISBN-10).

- Old behavior also never validated checksums (length 10/13 only), so this
  isn't a regression — but it does codify "untrusted identifiers go into the
  exported finder fields."
- The risk: a typo'd checksum gets carried into exports and will drive
  Hardcover lookups, and later edition *creation*/dedup on a bad value.
  The raw `isbn` field is always kept on the record for review, so dropping
  bad values from `isbn_10`/`isbn_13` loses no information.
- The package is already conservative where it matters: the derived
  counterpart (`Result.ISBN10()`/`ISBN13()`) is only filled on a valid
  checksum.

My recommendation: gate `Parse`/`Split` on a valid checksum too, and document
it — the record still keeps the raw `isbn`, so the value isn't lost, and the
export only carries identifiers you'd trust a lookup against. Happy to be
overruled if there's a reason ABS routinely serves non-checksum-valid ISBNs
(placeholders, etc.); either way, please make the policy explicit in the
`isbn` package docs since `TestAddWithMetadata_ISBNForms` currently pins the
permissive behavior.

### Question 2 — ebook + non-audiobook format leak (minor)
In `ToEditionExport` the ebook branch only overrides `EditionFormat` when it
is `""` or `"Audiobook"`. Production-created records always go through
`MarkEbook()`, so they export `"Ebook"` — but a record carrying
`reading_format: "ebook"` and a platform label like `"Audible Audio"`
(legacy, imported, or hand-built) would keep the audiobook platform hint.
Consider forcing `"Ebook"` for any ebook record as defense in depth.

### Action — update the Audiobookshelf schema/fixture
The checked-in ABS OpenAPI schema (`internal/api/audiobookshelf/
audiobookshelf-openapi.json`) documents `bookMetadataBase` with
`... isbn, asin, language, explicit` but **not `abridged`**, even though the
sibling `explicit` flag is there and live ABS responses emit both. Since this
PR introduces the decode contract for `abridged`, we should update the schema
(add `abridged` beside `explicit` in `bookMetadataBase`) and, ideally, the raw
response fixture in the same directory. It's doc-only, but per the repo's
"inspect the schema before changing an API response shape" rule, the schema
should reflect what the client now decodes.

### Nits
- `TestSavedExportKeepsTheCanonicalReadingFormatAndNoPlatformInfo`
  (`internal/mismatch/reading_format_test.go`) indexes `files[0]`/`files[1]`
  from `filepath.Glob` output and relies on order. If glob ordering isn't
  guaranteed on all filesystems, this can flake on CI — consider matching
  each file by content (e.g., find the one with `reading_format: "ebook"`)
  or sorting explicitly.
- The changelog bullet is missing the PR number; other bullets in the same
  section carry `(#NNN)`. Add `(#195)` for consistency.
- PR description says ebooks export with "no audio length", but the export
  writes `"audio_seconds": 0` (the changelog says 0 — the description is the
  one that's slightly off).
- `isbn.Parse`/`Result`/`ISBN10()`/`ISBN13()` have no production caller in
  this PR (only `Split` is used). Acknowledged foundation for step 2 — just
  flagging so the next step lands before this rots.

Overall: ship after the ISBN policy decision, the schema update, and a CI
rerun.

Snuffy2:
Thanks for the careful review, and for checking the callers and the importer. I've pushed fixes for the points below.

### Question 1: checksum-invalid ISBNs

I kept accepting an ISBN by shape and kept the value in `isbn_10` / `isbn_13`, and added a way to see whether it is valid instead of dropping it:

- `isbn.Result` now has a `Valid` field (the input's own check digit), and the `isbn` package docs state the policy: accept by shape, report the checksum, and derive the other form only from a valid ISBN-10 or a valid 978 ISBN-13.
- The export adds `isbn_10_valid` / `isbn_13_valid`, the same names Hardcover uses on `editions`. Each is omitted when that ISBN is empty, so `false` never means "no ISBN". The `edition` command decodes with a plain `json.Unmarshal`, so it ignores them.
- `TestParse` pins `Valid`, and `TestToEditionExport_ISBNChecksumFlags` checks valid, 979, bad-checksum, and empty values in the exported JSON.

Why I didn't gate on the checksum here: Hardcover's docs list `isbn_10_valid` / `isbn_13_valid` as flags on `editions`, which suggests it stores such ISBNs and flags them rather than rejecting them, and the sync matching in step 5 needs to find an edition by the value as given. The raw `isbn` is kept either way, so nothing is lost.

I want to be upfront about what I haven't verified. I haven't tried `insert_edition` or an ISBN lookup with a bad checksum against the real API yet, and I don't know how often Audiobookshelf serves non-checksum-valid ISBNs. So this is a reasoned choice, not a tested one. I will test before Step 3/4.

Your concern is real for creating editions: a typo'd ISBN would create an edition Hardcover flags as invalid. I've queued that decision for the draft and create endpoints (steps 3 and 4), where the flag will be available. That is where I'd decide between warning and rejecting, after checking the real API.

### Question 2: ebook with an audiobook platform label

Fixed. `ToEditionExport` now forces `Ebook` for any ebook record, replacing whatever label the record carries, which matches R11. `TestEbookExportIgnoresAnAudiobookFormatLabel` covers `""`, `Audiobook`, `Audible Audio`, and `libro.fm`.

### Audiobookshelf schema

Added `abridged` beside `explicit` in `bookMetadataBase` in `audiobookshelf-openapi.json`.

I didn't change `audiobookshelf_raw_response.json`. It is a 129-byte file that the client's debug logging writes at runtime, no test reads it, and it has neither `isbn` nor `explicit` either, so there was no fixture shape to extend. Let me know if you meant a different file.

### Nits

- `TestSavedExportKeepsTheCanonicalReadingFormatAndNoPlatformInfo` now sorts the files explicitly. `filepath.Glob` output was already in lexical order and the names carry a `%03d` sequence prefix, so the flake risk was low, but the test no longer relies on it.
- Added `(#195)` to the CHANGELOG bullet, which also now mentions the two new keys.
- Corrected the description: ebooks export `audio_seconds: 0`, not "no audio length".
- `isbn.Parse`, `Result`, `ISBN10()`, and `ISBN13()` have no production caller in this PR, as noted. The edition creator hardening (step 2) is the next PR and is the first caller.

One behavior note I've added to the description: because of the new flags, the audiobook export is no longer strictly unchanged. Any export that has an ISBN now also carries one of the `isbn_*_valid` keys. It is additive, and I've called it out in the CHANGELOG.

drallgood (approved):
Thanks for working through this so quickly — I've checked the new commits and
everything matches what you describe.

- The `Result.Valid` policy (accept by shape, report the checksum, derive the
  counterpart only from a valid ISBN-10 / valid 978 ISBN-13) is exactly the
  conservative middle ground, and `TestToEditionExport_ISBNChecksumFlags`
  covering the `false`-present, `979`, empty, and non-ISBN cases is the right
  shape.
- Forcing `Ebook` for any ebook record closed the platform-label leak; the
  `TestEbookExportIgnoresAnAudiobookFormatLabel` table is good.
- `abridged` next to `explicit` in `bookMetadataBase` and the explicit
  `sort.Strings(files)` both landed as suggested.

One residual item, which you've already flagged yourself: `isbn_10_valid` /
`isbn_13_valid` are not present in the checked-in Hardcover schema in this
repo (only `isbn_10`/`isbn_13` are), so their existence on `editions` rests on
the hosted API. Since the export is self-consistent and the tool ignores the
keys this is fine to merge, but please confirm the field names and behavior
against the real API before steps 2–5 rely on them, exactly as you planned.

I'm satisfied with this. Approve pending sync with develop.

Snuffy2:
Following up on the `isbn_10_valid` / `isbn_13_valid` question: I checked against the real Hardcover API, first read-only and then with writes. The results and what I'm doing about them:

### Confirmed
- **The fields exist.** `isbn_10_valid` and `isbn_13_valid` are booleans on `editions`. They are `null` when the ISBN is empty, which matches our export omitting the key. On 27 sampled ISBN-10 and ISBN-13 values (valid, invalid, 978 and 979), Hardcover's flag matched `internal/isbn`'s checksum every time.
- **Bad checksums are stored, not rejected.** Creating an edition with the hyphenated ISBN-13 `978-0-306-40615-8` stored `9780306406158` with `isbn_13_valid: false`. So Hardcover also strips hyphens on write, and an exact lookup only matches the stored, hyphen-free value.
- **Hardcover derives the other ISBN form.** A valid ISBN-10 on its own also stored the matching ISBN-13, flagged valid, and it is the same value our package derives. Nothing was derived for the bad-checksum ISBN-13. So a created edition can hold one form or both, and step 5 still must not depend on both.
- **Ebook editions.** `reading_format_id` 4 and the `Ebook` label are stored as sent.
- **Duplicate detection works.** Re-running the audiobook input made our creator find the existing edition by ASIN and return `existing: true`, with no second edition.
- **Language and country.** `language_id` 1 and `country_id` 1 are English and the United States.

### Differences from what I had assumed
- **Creating an edition needs a token scope.** A token without `write:catalog:append` gets `403 insufficient_scope` ("Missing scopes: write:catalog:append") and nothing is created. This is not part of the sync token's scopes, so the profile's token needs it for the add-edition feature. Step 2 documents the scopes the key needs in `cmd/edition/README.md` (`read:library`, `write:library`, `read:catalog` and `write:catalog:append`).
- **Format label normalization.** I sent `Audible Audio` and Hardcover stored `Audible`. I'm leaving the label as it is and noting it in the crosswalk. The preview will show what we send, not necessarily what gets stored.
- **Cover upload did not work with an API token.** The creator's request to `hardcover.app/api/upload/google` answered `401` to a new scoped token, and that endpoint is not part of Hardcover's documented API. The edition was still created. Hardcover attached a cover to the ebook edition by itself from its ISBN. Tokens from before August 2026 (all-access) may still work; I did not test one. I have not tried whether `insert_image` accepts an external URL. So nothing uploads a cover for now, and nothing tries: the creator's cover code stays in the tree but is switched off (`Creator.EnableCoverUpload`, which no caller uses). The `edition` command reports a requested `image_url` in `image_error` without fetching it, and `image-tool` reports that upload is not supported yet.

### Decisions
1. **Step 1: no code change.** The flags exist and behave as our export assumes. The checked-in schema snapshot still lacks them; I haven't refreshed it.
2. **Step 2:** the scopes are documented, the creator's cover upload is switched off (the PNG or JPEG only and at most 15 MiB rules stay in the code and are tested with it switched on, but none of the upload path was verified live), and the redirect token leak is fixed.
3. **No cover upload from anywhere, for now.** The app and the `edition` and `image-tool` commands attempt none. Steps 3, 4 and 7 drop the cover URL, the `image_url` and the cover warning.
4. **Steps 3, 4 and 7: a scope gate.** A token's scopes can't be read: there is no introspection endpoint, the token is opaque, and no schema field exposes them. Instead step 4 adds a pre-flight check that calls the edition insert with a book id that cannot exist. A 200 means the token has the scope, and a 403 names the missing scope. It creates nothing, and I confirmed both outcomes with two tokens. The backend exposes the result per profile, and step 7 shows the add-edition button only when the check passes.
5. **Step 5:** no change. Matching keeps working by whichever ISBN form the edition holds.

drallgood:
@Snuffy2 Thanks for the investigation.

I think the cover upload is broken right now due to changes on the Hardcover side. 
I will open an bug for that (indeed it's using private APIs so that's to be expected)

PR #197, opened 2026-09-22
feat: harden the edition creator (duplicate detection, formats, token handling)

Snuffy2:
## Summary of Changes

Hardens the standalone edition-creation workflow so it can safely reuse an existing Hardcover edition or create the requested audiobook or ebook edition without risking a duplicate after a failed lookup.

- Reuse same-book editions found by ASIN, ISBN-13, ISBN-10, or converted ISBN forms, while refusing editions owned by a different book.
- Distinguish explicit not-found responses from transport, GraphQL, and malformed-response failures; only explicit misses may proceed to `insert_edition`.
- Honor the requested edition format and optional `reading_format`, including ebook-specific payload behavior.
- Keep cover upload disabled while preserving guarded PNG/JPEG validation, the 15 MiB download limit, and failure reporting through `image_error`.
- Restrict Audiobookshelf credentials to the configured HTTPS origin and canonical base path, including redirect, port, traversal, subdomain, and downgrade handling.
- Document the local config-file workflow and required Hardcover token scopes.

## Multi-Step Project

1. ~~ISBN and export foundations: Add shared ISBN normalization and reading-format helpers, and fix the mismatch export (hyphenated ISBNs are kept, no default publisher, ebook items export as ebook editions, an abridged audiobook exports as Abridged).~~

2. Edition creator hardening (this PR): Make the edition creator reuse an existing edition of the same book by ASIN or ISBN and refuse another book's, honor the requested edition format, send the Audiobookshelf token only to its own server, keep the cover upload code but switched off, and create ebook editions.

3. Edition draft endpoint: Add a read-only endpoint that drafts a new Hardcover edition from a needs-review book's Audiobookshelf item, with author, narrator, and publisher resolved.

4. Edition create endpoint: Add the endpoint that creates the drafted edition on Hardcover, with request validation, a per-book in-flight guard, shutdown draining, and dry-run support.

5. Sync identifier matching: Match an existing edition by ASIN, ISBN-13, or ISBN-10 (either form) during sync so the book no longer lands in Needs review.

6. Immediate read-status resync: After an edition is created, re-sync that one book's read status right away, without overlapping a full sync and without waiting for the next one.

7. Sync Status UI: Add an "Add edition to Hardcover" action on needs-review records, with a preview, confirmation, and a resync option.

[Full Plan Document](https://github.com/Snuffy2/audiobookshelf-hardcover-sync/blob/docs/needs-review-edition-plan/docs/implementations/needs-review-edition-creation.md)

## Testing Instructions

- `GOTOOLCHAIN=go1.26.8 go test ./...`
- `GOTOOLCHAIN=go1.26.8 go test -race -count=1 ./internal/edition ./internal/api/hardcover`
- `GOTOOLCHAIN=go1.26.8 make lint`
- `GOTOOLCHAIN=go1.26.8 make build`
- Hosted tests, Go 1.26 tests, filesystem scan, container scan, and Trivy scan all pass.
- To exercise the command manually, run `edition create` twice with the same configured input and confirm the second result reports `"existing": true` without another mutation. Repeat with `"reading_format": "ebook"` to verify ebook payload behavior.
- Supplying `image_url` should still create the edition, make no cover request, and report that cover upload is unavailable through `image_error`.

## Checklist

- [x] Tests pass locally
- [x] Tests pass on CI
- [x] Changelog is updated


drallgood (changes requested):
## Review: feat: harden the edition creator

Nice, focused hardening. The deduplication logic, the fail-closed lookup policy, and the redirect/token scoping are all well designed and — unusually — actually well tested (the redirect and scoping tables cover scheme downgrade, port, subdomain, path-prefix, and even `%2e%2e` traversal). The `EditionResult.ImageError` convention of fixed, non-remote strings is the right call.

A few points, roughly in priority order:

**1. `InsecureSkipVerify: true` should be fixed in this PR (creator.go:246).** Pre-existing, but it is the HTTP client that carries both the Audiobookshelf token and the Hardcover token, and it makes the otherwise excellent token scoping trivially bypassable by any on-path attacker (including, notably, a 5-minute CLI run against a split/controlled host). Even if full verification is a follow-up, flipping the default to `false` and gating the insecure behavior behind a clearly-opt-in call would materially improve this PR's security posture.

**2. The legacy token fallback is the weakest part of the scoping story (creator.go:183–185).** `strings.Contains(imageURL, "audiobookshelf")` will happily send the bearer token to e.g. `https://attacker.example/hosted-root/audiobookshelf/x` whenever no base URL is configured. Since the commands always call `SetAudiobookshelfBaseURL`, this only bites with an empty config URL — but consider dropping the raw substring heuristic entirely, or tightening it (https-only plus a hostname check) so the default isn't an open substring match.

**3. Two fail-safe edge cases worth a README note:**
- With a base URL configured, the token is withheld from the *same* server when the image URL uses a different hostname/port (e.g. `localhost` vs. the configured host). Security — fine — but users with non-canonical ABS addresses will silently stop getting authenticated cover fetches.
- A base URL without a scheme (`abs.home:13378`) parses with `abs.home` as scheme, leaving `Host` empty, so tokens are silently never sent. A tiny normalization/validation in `SetAudiobookshelfBaseURL` would prevent this footgun.

**4. `image-tool` is now unconditionally non-functional by design.** Both upload subcommands exit(1) with a generic "Failed to upload …" message. That's documented, but a dedicated, explicit message (and ideally a non-error exit or a clear "not supported" banner) would be much kinder than routing `ErrCoverUploadDisabled` through the generic failure path.

**5. Minor / informational:**
- The redirect policy is installed on the shared client, so it governs the `hardcover.app/api/upload/google` and GCS requests too — safe (credentials get stripped on any redirect there), but if that credential endpoint ever redirects, it will silently break while covers are disabled. Worth a comment or scoping the policy to the download request.
- A book that exists on Hardcover **only** in the other reading format now fails `findExistingEdition` with "book has no editions" instead of proceeding (the queries are format-scoped). Probably the intended strictness; the changelog entry could call it out explicitly.
- Low-priority: `http.DetectContentType` sniffs only the first 512 bytes, so a JPEG-prefixed HTML "polyglot" would still pass the format gate; the existing headers/magic-byte checks are already a big improvement, so this is just a note.

Overall: solid work. The duplicate-detection and credential-scoping changes are well-tested and I'd be comfortable merging after the TLS-default question is answered (or explicitly deferred with an issue).

Snuffy2:
Thanks for the detailed review. I’ve rechecked each point against the current PR head:

1. **TLS verification:** The creator now verifies certificates by default. Insecure TLS requires an explicit `EnableInsecureTLS` call, which no production caller makes.

2. **Token fallback:** I removed the substring-based fallback. With no configured Audiobookshelf base URL, the token is withheld. Otherwise, the URL must match the configured scheme, host, port, and canonical path.

3. **Base-URL edge cases:** The command README explains that aliases and different ports do not receive the token. An unambiguous bare address such as `abs.home:13378` is normalized to HTTPS; ambiguous forms require an explicit scheme.

4. **`image-tool` UX:** Both upload paths now give an explicit “cover uploads are not supported yet” message. Help and documentation also state that uploads are disabled. The command still exits nonzero because no upload occurred.

5. **Shared redirect policy:** Confirmed. The code now documents that it also governs the retained Hardcover credential and GCS requests. A redirected credential request can lose its Authorization header; this should be revisited before cover uploads are enabled. The GCS request uses signed form fields, not that header.

6. **Other-reading-format lookup:** I could not confirm the reported failure path. The query filters books and editions by the requested format, so a book with only the other format is treated as not found and creation proceeds. “Book has no editions” indicates an inconsistent response to that filtered query, where failing closed is intentional.

7. **JPEG-prefixed polyglot:** Confirmed as a limitation: `http.DetectContentType` checks a signature, not whether the full file decodes as an image. Cover uploads remain disabled in production. Proper JPEG/PNG decoding and a regression test should be added before enabling that retained upload path.

drallgood (approved):
Verified all six points against the current head — all look good.

1. **TLS:** Default is now Go-verified; `EnableInsecureTLS` is an explicit opt-in with no production caller and a pinning test. Resolved.
2. **Fallback:** Confirmed removed — empty base withholds the token; only a scheme+host+port+canonical-path match sends it. Resolved.
3. **Base-URL forms:** The validation + normalization + error propagation is a clean resolution (bare `host:port` → https, ambiguous forms rejected loudly). README + `TestSetAudiobookshelfBaseURLValidation` cover it. Resolved.
4. **image-tool:** The dedicated "cover uploads are not supported yet" message and docs (with nonzero exit retained and explained) resolve the UX point.
5. **Shared redirect policy:** Thanks for documenting the hardcover.app/GCS implications and the signed-form-field detail; the defer-to-re-enable pointer is the right one.
6. **Other-format lookup:** Confirmed — the format filter sits in the book-level `where`, so an other-format-only book is a clean not-found and creation proceeds; "no editions" on a matched book is an inconsistent response where fail-closed is correct. My original concern was overstated; conceded.
7. **Polyglot:** Agreed — signature vs. full decode; the regression test + real decoding plan for the retained path is the right gate before enabling.

No remaining blockers from my side. The `InsecureSkipVerify` removal in particular addresses the biggest concern.

